PR #4, opened 2021-11-13
Fixes #3; Extension send transaction window content does not fill space.

### LINKED ISSUE #3: Extension send transaction window content does not fill space
Hi, I noticed that when I send a transaction the window content does not fill all available space. I use a tiling window manager so the window is by default bigger than the default size. 

It would be nice if the content fills the extra space, I attached a screenshot where you can see the problem.

2021-11-12-205850_3840x1080_scrot

graemeblackwood:


janek26:
thanks

PR #6, opened 2021-11-16
feat: request connection

janek26:
https://argent.atlassian.net/browse/BLO-12

We need to be carefull not to expose the provider in window to every app. We should check how MetaMask does that as they had to change that specific behaviour.

Upon connecting the wallet from a dapp, the extension should open and request to connect to that specific dapp. The provider with the account, etc should only be injected if approved. The extension should keep a mapping of the approved dapps.

This MR also improves encryption/decryption when development mode is detected

### DIFF
--- a/packages/extension/src/app/screens/Connect.tsx
+++ b/packages/extension/src/app/screens/Connect.tsx
@@ -1,0 +1,27 @@
+import { FC } from "react"
+import styled from "styled-components"
+
+import { P } from "../components/Typography"
+import { Confirm, ConfirmPageProps } from "./Confirm"
+
+interface ConnectProps extends ConfirmPageProps {
+  host: string
+}
+
+const Code = styled.code`
+  background-color: rgba(255, 255, 255, 0.15);
+  border-radius: 2px;
+  padding: 0 0.5em;
+`
+
+export const ConnectScreen: FC<ConnectProps> = ({ host, ...props }) => {
+  return (
+    <Confirm title="Connect to DApp" confirmButtonText="Connect" {...props}>
+      <P>
+        <Code>{host}</Code> tries to connect to your wallet. If you allow this
+        request the website will be able to read you wallet addresses and
+        request transactions, which you still need to sign.
+      </P>
+    </Confirm>
+  )
+}
--- a/packages/extension/src/app/states/RouterMachine.ts
+++ b/packages/extension/src/app/states/RouterMachine.ts
@@ -43,7 +43,7 @@ type RouterEvents =
   | { type: "APPROVED_TX" }
   | { type: "GENERATE_L1"; data: { password: string } }
 
-type Context = {
+interface Context {
   password?: string
   selectedWallet?: string
   l1?: ethers.Wallet
--- a/packages/extension/src/app/states/RouterMachine.ts
+++ b/packages/extension/src/app/states/RouterMachine.ts
@@ -7,6 +7,7 @@ import { DoneEvent, assign, createMachine } from "xstate"
 import {
   getLastSelectedWallet,
   messenger,
+  readPendingWhitelist,
   readRequestedTransactions,
 } from "../utils/messaging"
 import { addToken } from "../utils/tokens"
@@ -23,8 +24,10 @@ type RouterEvents =
   | { type: "REJECT_TX" }
   | { type: "SHOW_ACCOUNT_LIST" }
   | { type: "SHOW_ADD_TOKEN" }
+  | { type: "SHOW_SETTINGS" }
   | { type: "FORGOT_PASSWORD" }
   | { type: "AGREE" }
+  | { type: "REJECT" }
   | { type: "SUBMIT_KEYSTORE"; data: string }
   | { type: "SELECT_WALLET"; data: string }
   | {
@@ -50,6 +53,7 @@ type Context = {
   txHash?: string
   wallets: Record<string, Wallet>
   isPopup?: boolean
+  hostToWhitelist?: string
   error?: string
 }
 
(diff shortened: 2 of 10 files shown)


packages/extension/src/app/states/RouterMachine.ts
```diff
@@ -23,8 +24,10 @@ type RouterEvents =
   | { type: "REJECT_TX" }
   | { type: "SHOW_ACCOUNT_LIST" }
   | { type: "SHOW_ADD_TOKEN" }
+  | { type: "SHOW_SETTINGS" }
   | { type: "FORGOT_PASSWORD" }
   | { type: "AGREE" }
+  | { type: "REJECT" }
   | { type: "SUBMIT_KEYSTORE"; data: string }
   | { type: "SELECT_WALLET"; data: string }
```
delaaxe, line 32: Isn't `payload` the preferred field name in React state machine events?

packages/extension/src/app/states/RouterMachine.ts
```diff
@@ -50,6 +53,7 @@ type Context = {
   txHash?: string
   wallets: Record<string, Wallet>
   isPopup?: boolean
+  hostToWhitelist?: string
   error?: string
 }
```
delaaxe, line 58: Prefer using `interface` whenever possible?

packages/extension/src/app/states/RouterMachine.ts
```diff
@@ -345,8 +378,16 @@ export const routerMachine = createMachine<
...
           const password = ctx.password!
           if (password) {
-            const backup = await ctx.l1!.encrypt(password, {}, (progress) =>
-              useProgress.setState({ progress, text: "Encrypting..." }),
+            const backup = await ctx.l1!.encrypt(
+              password,
+              {
+                scrypt: {
+                  // The number must be a power of 2 (default: 131072)
+                  N: isDev ? 64 : 131072,
```
delaaxe, line 386: Great

packages/extension/src/background.ts
```diff
@@ -169,6 +214,35 @@ browser.runtime.onConnect.addListener(function (port) {
       case "RESET_ALL": {
         return browser.storage.local.clear()
       }
+
+      case "ADD_WHITELIST": {
+        return addPendingWhitelist(data)
+      }
```
delaaxe, line 220: Braces shoudn't be needed around `case`s??
```suggestion
      case "ADD_WHITELIST":
        return addPendingWhitelist(data)
```

packages/extension/webpack.config.js
```diff
@@ -7,6 +7,8 @@ const htmlPlugin = new HtmlWebPackPlugin({
   filename: "./index.html",
 })
 
+const isProd = process.env.NODE_ENV === "production"
```
delaaxe, line 10: Should it be colocated with `isDev`?

delaaxe (approved):

packages/extension/src/app/states/RouterMachine.ts
janek26: yep, we could change it in a future PR, but this PR isn't touching these parts

packages/extension/src/app/states/RouterMachine.ts
janek26: thanks, hope it helps :)

packages/extension/src/background.ts
janek26: that's right, but I prefer them. 
JS and TS are curly brace languages and I prefer to use them over indentations, which is an anti pattern in JS.
it also helps to not forget return, as it'd otherwise execute multiple cases :)

packages/extension/webpack.config.js
janek26: disagree, as those 2 expressions run in very different contexts. 
This expression is run by Webpack as part of evaluating the Webpack config, and sources it's process.env from the actual process var, which is available in node context.

The isDev expression from above works completely different! It gets replaced (!) by Webpack, as process isn't valid in a browser context, therefore the expression needs to look exactly like it does, and these both processes shouldn't share any files.

packages/extension/src/background.ts
delaaxe: It's fine, it's just that I had never seen this pattern in any other curly brace language project in the past, so I was a bit surprised as to the reason. I thought it was a workaround for the lack of semicolons

packages/extension/webpack.config.js
delaaxe: Oh right, nevermind then

delaaxe (approved):

packages/extension/src/app/screens/Connect.tsx
```diff
@@ -0,0 +1,27 @@
...
+  background-color: rgba(255, 255, 255, 0.15);
+  border-radius: 2px;
+  padding: 0 0.5em;
+`
+
+export const ConnectScreen: FC<ConnectProps> = ({ host, ...props }) => {
+  return (
+    <Confirm title="Connect to DApp" confirmButtonText="Connect" {...props}>
+      <P>
+        <Code>{host}</Code> tries to connect to your wallet. If you allow this
```
juniset, line 21: I would use "wants" instead of "tries" as it sounds less sneaky.

packages/extension/src/background.ts
delaaxe: Also there might be an eslint rule that can help you not forget return statements in switch cases

packages/extension/src/app/states/RouterMachine.ts
juniset, line 386: I think for now we can go from 2^17 (131072) to 2^15 (32768), which will cut the decryption time by ~4.

juniset (approved):

janek26:
thanks @juniset 
as it was merged before your review I added the changes to `main`

PR #9, opened 2021-11-18
Add a list of default tokens

delaaxe:
Add a list of basic, "well known" ERC20 tokens in the extension's code to avoid showing an empty wallet when a user recovers an existing wallet.

Might need to check how often token details are cached and refreshed as this list grows in size.

### DIFF
--- a/packages/extension/src/app/screens/Account.tsx
+++ b/packages/extension/src/app/screens/Account.tsx
@@ -86,9 +86,9 @@ const TokenList: FC<{
         <TokenListItem
           key={i}
           index={i}
-          decimals={token.decimals!.toNumber()}
+          decimals={token.decimals?.toNumber()}
           balance={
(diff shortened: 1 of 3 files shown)


packages/extension/src/app/screens/Account.tsx
```diff
@@ -86,9 +86,9 @@ const TokenList: FC<{
         <TokenListItem
           key={i}
           index={i}
-          decimals={token.decimals!.toNumber()}
+          decimals={token.decimals?.toNumber()}
```
delaaxe, line 89: The amount of bugs caused by the use of the non-null assertion operator is so large it's not even funny. I personally never ever use it.

I propose adding an eslint rule to forbid it's usage.

janek26: good idea, I usually also try to avoid it.. To avoid it starknet.js would also need a types upgrade

janek26 (approved):

PR #12, opened 2021-11-18
feat: alpha 4

janek26:


packages/extension/src/app/Wallet.ts
```diff
@@ -111,7 +111,7 @@ export class Wallet {
     }
 
     return new Wallet(
-      deployTransaction.address,
+      deployTransaction.address!,
```
janek26, line 114: Hat to See it too, but wit starknet not returning the address on an AddTransaction object every time, I think it's the best solution.
.address will be always there, because the transaction referenced will always be a deploy tx. just invokeFunction tx's do not have this property.
I guess this needs to be more specific in starknet.js

delaaxe (approved):

janek26:
thanks! missed some packages

PR #18, opened 2021-11-26
Feature/move key to bg

### LINKED ISSUE #5: Decrypting wallet password takes a long time
Screen Shot 2021-11-14 at 7 34 47 AM

Argent X uses `ethers.js` to encrypt and decrypt passwords. This specific function is taking a long time to decrypt:

```javascript
          const l1 = await ethers.Wallet.fromEncryptedJson(
            keyStore,
            event.data.password,
            (progress) =>
              useProgress.setState({ progress, text: "Decrypting ..." }),
          )
```

Which really reduces the UX for the user.

### Possible solutions 
1. According to the `ethers.js` docs [here](https://docs.ethers.io/v5/concepts/security/#security--security--pbkdf--work-arounds-not-recommended), we can fix it by changing `scrypt.N value` which is not recommended.

2. We could reduce the times that Argent X asks the user to decrypt the password.

3. We can use MetaMask's [encrypt/decrypt function](https://github.com/MetaMask/browser-passworder) - though I'm not sure why they don't believe long decryption times is a feature and not a bug. More details on MetaMask security [here](https://www.wispwisp.com/index.php/2020/12/25/how-metamask-stores-your-wallet-secret/)

### LINKED ISSUE #8: Request: auto-lock options
The wallet is auto-locked when the popup loses focus. For convenience there could be an option to toggle auto-lock or specify an amount of time.

janek26:
closes #5 
closes #8


packages/extension/src/background/session.ts
```diff
@@ -0,0 +1,42 @@
+import { getL1 } from "./keys/l1"
+
+let sessionPassword: string | undefined
+let sessionStartTime: number | undefined
+let sessionTimeoutDelay: number | undefined
+let sessionTimeout: number | undefined
+
+export function startSession(
+  password: string,
+  duration: number = 15 * 60 * 60 * 1000,
```
delaaxe, line 10: Could use variable name to make duration explicit

packages/extension/src/ui/App.tsx
```diff
@@ -121,7 +121,7 @@ function App() {
   if (state.matches("settings"))
     return <Settings onBack={() => send("GO_BACK")} />
 
-  if (state.matches("account"))
+  if (state.matches("account")) {
```
delaaxe, line 124: I would add a prettier/eslint that automatically inserts curly brackets when they're missing. This is a common source of bugs

delaaxe (approved):

packages/extension/src/ui/App.tsx
janek26: prettier doesn't allow this.. eslint does for sure :) I'll create a follow up ticket

PR #19, opened 2021-11-29
feature: network switcher

delaaxe:
Adds a network switcher to the account page and uses the corresponding network in wallet & page contents

### DIFF
--- a/packages/extension/package.json
+++ b/packages/extension/package.json
@@ -41,7 +41,7 @@
     "react-hook-form": "^7.19.4",
     "react-promise-suspense": "^0.3.3",
     "react-router-dom": "^6.0.1",
-    "starknet": "seanjameshan/starknet.js#develop",
+    "starknet": "^2.1.0",
     "styled-components": "^5.3.3",
     "styled-normalize": "^8.0.7",
     "webextension-polyfill": "^0.8.0",
--- a/packages/extension/package.json
+++ b/packages/extension/package.json
@@ -41,7 +41,7 @@
     "react-hook-form": "^7.19.4",
     "react-promise-suspense": "^0.3.3",
     "react-router-dom": "^6.0.1",
-    "starknet": "^2.0.1",
+    "starknet": "seanjameshan/starknet.js#develop",
     "styled-components": "^5.3.3",
     "styled-normalize": "^8.0.7",
     "webextension-polyfill": "^0.8.0",
--- a/packages/extension/package.json
+++ b/packages/extension/package.json
@@ -41,7 +41,7 @@
     "react-hook-form": "^7.19.4",
     "react-promise-suspense": "^0.3.3",
     "react-router-dom": "^6.0.1",
-    "starknet": "^2.1.0",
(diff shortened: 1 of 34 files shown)


juniset:
When I switch to mainnet I land on the Account screen with the possibility to create a new account. There is no way to cancel the action and go back (I'm forced to create an account). We should add a back arrow in the top left corner or a cancel button in the bottom.

delaaxe:
> When I switch to mainnet I land on the Account screen with the possibility to create a new account. There is no way to cancel the action and go back (I'm forced to create an account). We should add a back arrow in the top left corner or a cancel button in the bottom.

But if you click on those, to which wallet should it bring you back? My thought was adding the network switcher to the account list page too

packages/extension/package.json
```diff
@@ -41,7 +41,7 @@
     "react-hook-form": "^7.19.4",
     "react-promise-suspense": "^0.3.3",
     "react-router-dom": "^6.0.1",
-    "starknet": "^2.0.1",
+    "starknet": "seanjameshan/starknet.js#develop",
```
janek26, line 44: should this go to main branch? 
No other way for now, as starknet.js deployment is still blocked, but I really don't like this :/ can we maybe create an issue to replace this ASAP ?

packages/extension/src/background/keys/l1.ts
```diff
@@ -137,10 +120,25 @@ function downloadTextFile(text: string, filename: string) {
   })
 }
 
-export const getWallets = () => store.getItem("walletAddresses")
+export const getWallets = async (networkId?: string): Promise<BackupWallet[]> =>
+  (await store.getItem("walletAddresses"))
+    .map((wallet) => {
+      if (typeof wallet === "string") {
+        // backwards compatibility
```
janek26, line 127:
```suggestion
        // TODO: remove backwards compatibility
```

packages/extension/src/inpage/index.ts
```diff
@@ -70,22 +72,37 @@ const starknetWindowObject: StarknetWindowObject = {
...
 window.starknet = starknetWindowObject
 
+window.addEventListener(
+  "message",
+  ({ data }: MessageEvent<WindowMessageType>) => {
+    const { starknet } = window
+    if (starknet && starknet.signer && data.type === "CHANGE_NETWORK") {
+      console.warn("CHANGE_NETWORK", data.data)
+      starknet.provider = new Provider({ network: data.data as any })
+      starknet.signer = new WalletSigner(
```
janek26, line 82: should we also set `starknet.selectedAddress` (as it probably has changed) here?

packages/extension/package.json
delaaxe: OK let's wait then maybe

packages/extension/src/inpage/index.ts
delaaxe: also this should only be broadcasted to the page if there is an actual address...

janek26 (approved):

janek26:
great work @delaaxe !

PR #38, opened 2021-12-01
feat: support new erc20

janek26:
This PR implements the new ERC20 as defined in openzeppelins repo, with shortstring symbol and name aswell as uint256 for token balances.

(!) The contract abi parsing needs to be revisited at some point to make it easier

### DIFF
--- a/packages/extension/src/assets/erc20-tokens.json
+++ b/packages/extension/src/assets/erc20-tokens.json
@@ -1,6 +1,6 @@
 [
   {
-    "address": "0x00a45e3942b7a75983dea7afffda9304d0273773619d1e3d5eaa757d751bfaf3",
+    "address": "0x07394cbe418daa16e42b87ba67372d4ab4a5df0b05c6e554d158458ce245bc10",
     "name": "Playground Token",
     "symbol": "TEST",
     "decimals": "18"
--- a/packages/extension/src/ui/App.tsx
+++ b/packages/extension/src/ui/App.tsx
@@ -1,5 +1,7 @@
+import { BigNumber } from "@ethersproject/bignumber"
 import { useMachine } from "@xstate/react"
 import { Suspense } from "react"
+import { uint256 } from "starknet"
 import { createGlobalStyle } from "styled-components"
 import { normalize } from "styled-normalize"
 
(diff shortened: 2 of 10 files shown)


packages/extension/src/assets/erc20-tokens.json
```diff
@@ -1,6 +1,6 @@
 [
   {
-    "address": "0x00a45e3942b7a75983dea7afffda9304d0273773619d1e3d5eaa757d751bfaf3",
+    "address": "0x07394cbe418daa16e42b87ba67372d4ab4a5df0b05c6e554d158458ce245bc10",
```
delaaxe, line 3: On which network is this? Network should be in an extra field here IMO

packages/extension/src/ui/App.tsx
```diff
@@ -21,6 +23,13 @@ import { WelcomeScreen } from "./screens/WelcomeScreen"
 import { routerMachine } from "./states/RouterMachine"
 import { TokenDetails } from "./utils/tokens"
 
+function getUint256CalldataFromBN(bn: BigNumber) {
+  return {
+    type: "struct" as "struct",
```
delaaxe, line 28:
```suggestion
    type: "struct" as const,
```

packages/extension/src/ui/utils/tokens.ts
```diff
@@ -10,7 +10,7 @@ import { isValidAddress } from "./addresses"
 export const playgroundToken = erc20Tokens.find(
   ({ address }) =>
     address ===
-    "0x00a45e3942b7a75983dea7afffda9304d0273773619d1e3d5eaa757d751bfaf3",
+    "0x07394cbe418daa16e42b87ba67372d4ab4a5df0b05c6e554d158458ce245bc10",
```
delaaxe, line 13: This should depend on network

packages/playground/src/services/token.service.ts
```diff
@@ -1,14 +1,18 @@
...
 export const erc20TokenAddress =
-  "0x00a45e3942b7a75983dea7afffda9304d0273773619d1e3d5eaa757d751bfaf3"
+  "0x07394cbe418daa16e42b87ba67372d4ab4a5df0b05c6e554d158458ce245bc10"
 
 const mintSelector = stark.getSelectorFromName("mint")
 
 const transferSelector = stark.getSelectorFromName("transfer")
 
+function getUint256CalldataFromBN(bn: number.BigNumberish) {
+  return { type: "struct" as "struct", ...uint256.bnToUint256(bn) }
```
delaaxe, line 13:
```suggestion
  return { type: "struct" as const, ...uint256.bnToUint256(bn) }
```

delaaxe (approved):

janek26:
We should bind the token to a specific network as soon as Maine supports its first ERC20

PR #109, opened 2021-12-21
feat(local-network): initial implementation

delaaxe:
Rudimentary implementation of local network support with:

- Addition of a `localhost` network whose port number can be changed in settings.
- Backup of wallets on custom networks by allowing urls in lieu of network ids.
- Update of playground page for the case where there is no known token to play with.

/!\ This PR likely shouldn't be part of a release without proper handling of errors, checking existence of wallets on networks, etc beforehand. /!\

### DIFF
--- a/packages/extension/src/ui/screens/SuccessScreen.tsx
+++ b/packages/extension/src/ui/screens/SuccessScreen.tsx
@@ -33,7 +33,10 @@ export const SuccessScreen: FC<SuccessScreenProps> = ({
   return (
     <SuccessScreenWrapper>
       <Spinner size={92} />
-      <SuccessText href={`${explorerUrl}/tx/${txHash}`} target="_blank">
+      <SuccessText
+        href={explorerUrl && `${explorerUrl}/tx/${txHash}`}
+        target="_blank"
+      >
         Transaction was submitted
       </SuccessText>
     </SuccessScreenWrapper>
--- a/packages/playground/src/components/TokenDapp.tsx
+++ b/packages/playground/src/components/TokenDapp.tsx
@@ -35,13 +35,28 @@ export const TokenDapp: FC = () => {
     })()
   }, [transactionStatus, lastTransactionHash])
 
+  const network = networkId()
+  if (network !== "goerli-alpha" && network !== "mainnet-alpha") {
+    return (
+      <>
+        <p>
+          There is no demo token for this network, but you can deploy one and
+          add its address to this file:
+        </p>
+        <div>
+          <pre>packages/playground/src/token.service.ts</pre>
+        </div>
+      </>
+    )
+  }
+
   const handleMintSubmit = async (e: React.FormEvent) => {
     e.preventDefault()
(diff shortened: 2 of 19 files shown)


packages/extension/src/ui/screens/ConfirmScreen.tsx
```diff
@@ -63,7 +63,11 @@ export const ConfirmScreen: FC<ConfirmScreenProps> = ({
           src={getAccountImageUrl(selectedAccount.accountNumber)}
           disabled
         />
-        <NetworkSwitcher networkId={selectedAccount.networkId} disabled />
+        <NetworkSwitcher
+          networkId={selectedAccount.networkId}
+          port={0}
```
janek26, line 68: I think we should pass the correct port here for displaying purposes

packages/extension/src/ui/screens/SuccessScreen.tsx
```diff
@@ -33,7 +33,10 @@ export const SuccessScreen: FC<SuccessScreenProps> = ({
   return (
     <SuccessScreenWrapper>
       <Spinner size={92} />
-      <SuccessText href={`${explorerUrl}/tx/${txHash}`} target="_blank">
+      <SuccessText
+        href={explorerUrl && `${explorerUrl}/tx/${txHash}`}
+        target="_blank"
+      >
```
janek26, line 39: instead of showing a non working link we could just hide the whole text

packages/playground/src/components/TokenDapp.tsx
```diff
@@ -35,13 +35,28 @@ export const TokenDapp: FC = () => {
...
+  if (network !== "goerli-alpha" && network !== "mainnet-alpha") {
+    return (
+      <>
+        <p>
+          There is no demo token for this network, but you can deploy one and
+          add its address to this file:
+        </p>
+        <div>
+          <pre>packages/playground/src/token.service.ts</pre>
+        </div>
```
janek26, line 48: instead of referencing the code where to add default tokens, shouldn't we just suggest to add the token by using the UI?

janek26 (commented):
I haven't seen any changes to the background page, aren't there any changes necessary ?

delaaxe:
> I haven't seen any changes to the background page, aren't there any changes necessary ?

Not sure what you mean, there are a few changes in `background/`

packages/extension/src/ui/screens/SuccessScreen.tsx
janek26: forget about this! I switched it up with the UI, but it's example dapp!

janek26:
I See, thanks!

packages/playground/src/components/TokenDapp.tsx
delaaxe: not sure I understand, this is the token used by that dapp code, adding a token in wallet wouldn't change that

janek26 (approved):

PR #138, opened 2022-01-04
feat: add token from dapp

janek26:
As a DApp Developer I want to be able to add my token to the wallet extension programmatically.

### DIFF
--- a/packages/extension/src/inpage/index.ts
+++ b/packages/extension/src/inpage/index.ts
@@ -91,6 +91,7 @@ const starknetWindowObject: StarknetWindowObject = {
 
         return resolve()
       }
+      return reject(Error("Not implemented"))
     }),
   enable: () =>
     new Promise((resolve) => {
--- a/packages/extension/src/inpage/index.ts
+++ b/packages/extension/src/inpage/index.ts
@@ -24,12 +24,74 @@ function sendMessage(msg: MessageType): void {
   )
 }
 
+function waitForMsgOfType<
+  K extends MessageType["type"],
+  T extends { type: K } & MessageType,
+>(
+  type: K,
+  timeout: number,
+  predicate: (x: T) => boolean = () => true,
+): Promise<T extends { data: any } ? T["data"] : undefined> {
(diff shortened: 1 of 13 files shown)


packages/extension/src/inpage/index.ts
```diff
@@ -24,12 +24,75 @@ function sendMessage(msg: MessageType): void {
...
+}
+
 // window.ethereum like
 const starknetWindowObject: StarknetWindowObject = {
   signer: undefined,
   provider: defaultProvider,
   selectedAddress: undefined,
   isConnected: false,
+  request: (call) =>
+    new Promise(async (resolve, reject) => {
```
delaaxe, line 57: No need to use `new Promise` here

packages/extension/src/inpage/model.ts
```diff
@@ -2,7 +2,26 @@ import type { Provider, SignerInterface } from "starknet"
...
+  options: {
+    address: string // The hexadecimal StarkNet address of the token contract
+    symbol?: string // A ticker symbol or shorthand, up to 5 alphanumerical characters
+    decimals?: number // The number of asset decimals
+    image?: string // A string url of the token logo
+    name?: string // The name of the token - not in spec
+  }
+}
+
+export type RpcMessage = {
```
delaaxe, line 18: Interface?

delaaxe (approved):

packages/extension/src/inpage/model.ts
janek26: yes and no, normally I would prefer an interface, but when adding more RpcMessages we need to go back to a type, as it's an enum, ie:
```
export type RpcMessage = {
  type: "rpcMethod1",
  data: ...,
} | {
  type: "rpcMethod2",
  data ...,
}
```

delaaxe: I know that was the intention but I would only switch from `interface` to `union` once it's actually the case, so no premature optimization.

Those things become obvious once you integrate a linter in your workflow.

PR #149, opened 2022-01-07
feat: transaction queueing

janek26:
Allows wallets to queue multiple transactions into one block, by increasing the nonce optimistically.
If one transaction fails all subsequent transactions will also fail (nonce mismatch)

### DIFF
--- a/packages/extension/src/background/index.ts
+++ b/packages/extension/src/background/index.ts
@@ -1,5 +1,5 @@
 import { compactDecrypt } from "jose"
-import { Provider, ec, encode } from "starknet"
+import { Provider, ec, encode, number, stark } from "starknet"
 
 import { ActionItem } from "../shared/actionQueue"
 import { messageStream, sendMessage } from "../shared/messages"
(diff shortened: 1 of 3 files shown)


packages/extension/src/background/index.ts
```diff
@@ -210,8 +215,14 @@ async function main() {
               network: selectedWallet.network,
             })
 
-            const tx = await signer.addTransaction(transaction)
+            const nonce = await getNonce(signer)
+
+            const tx = await signer.addTransaction({
+              ...transaction,
+              nonce: nonce,
+            })
```
delaaxe, line 223:
```suggestion
            const tx = await signer.addTransaction({ ...transaction, nonce })
```

packages/extension/src/background/nonce.ts
```diff
@@ -0,0 +1,39 @@
+import { Signer, number, stark } from "starknet"
+
+const nonceStore: {
+  [walletAddress: string]: string
+} = {}
```
delaaxe, line 5:
```suggestion
const nonceStore: Record<string, string> = {}
```

delaaxe (approved):

packages/extension/src/background/nonce.ts
janek26: oh sorry, I don't get used to it! :D so much better

PR #186, opened 2022-01-20
feat: isolate token balance fetch

janek26:
This PR reduces the calls to the sequencer from 4 to 1 for each token to update the balance.
It also removes old clunky patterns and uses zustand instead


packages/extension/src/ui/states/tokens.ts
```diff
@@ -0,0 +1,141 @@
...
+                key,
+                (value as TokenDetails[]).filter(
+                  (token) =>
+                    !parsedDefaultErc20Tokens.some((defaultToken) =>
+                      equalToken(token, defaultToken),
+                    ),
+                ),
+              ]
+            }
+            return [key, value]
```
delaaxe, line 72: A bit spaghetti-ish

packages/extension/src/ui/states/tokens.ts
```diff
@@ -0,0 +1,141 @@
...
+
+export const addToken = (token: TokenDetails) => {
+  useTokens.getState().addToken(token)
+}
+
+export const selectorTokensByNetwork =
+  (networkId: string) => (state: TokenState) =>
+    state.tokens.filter((token) => token.networkId === networkId)
+
+export interface TokenDetailsWithBalance extends TokenDetails {
```
delaaxe, line 87: nice

packages/extension/src/ui/states/tokens.ts
```diff
@@ -0,0 +1,141 @@
...
+        ),
+    },
+  ),
+)
+
+export const addToken = (token: TokenDetails) => {
+  useTokens.getState().addToken(token)
+}
+
+export const selectorTokensByNetwork =
```
delaaxe, line 83:
```suggestion
export const selectTokensByNetwork =
```

packages/extension/src/ui/states/tokens.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  const { data, isValidating, error, mutate } = useSWR(
+    [walletAddress, ...tokenAddresses],
+    async (walletAddress, ...tokenAddresses) => {
+      const balances = await Promise.all(
+        tokenAddresses.map(async (address) =>
+          fetchTokenBalance(address, walletAddress, networkId),
+        ),
+      )
+      return balances.reduce((acc, balance, i) => {
```
delaaxe, line 109: Avoid using `reduce`...?

packages/extension/src/ui/utils/tokens.ts
```diff
@@ -1,80 +1,17 @@
...
-
-export interface TokenDetails {
-  address: string
-  name?: string
-  symbol?: string
-  decimals?: BigNumber
-  balance?: BigNumber
-  networkId: string
-}
-
```
delaaxe, line 77: good riddance, that whole part was a little clunky

packages/extension/src/ui/utils/tokens.ts
```diff
@@ -131,19 +68,29 @@ export const fetchTokenDetails = async (
...
+): Promise<BigNumber> => {
+  const provider = getProvider(networkId)
+  const tokenContract = new Contract(parsedErc20Abi as Abi[], address, provider)
+  const balance = await tokenContract
+    .call("balanceOf", { user: walletAddress })
+    .then((x) =>
+      BigNumber.from(uint256.uint256ToBN(x.balance as any).toString()),
+    )
+    .catch(() => BigNumber.from(0))
+  return balance
```
delaaxe, line 95: Callback based programming... no

```suggestion
  try {
    const result = await tokenContract.call("balanceOf", { user: walletAddress })
    return BigNumber.from(uint256.uint256ToBN(result.balance as any).toString())
  } catch {
    return BigNumber.from(0)
  }
```

delaaxe (approved):

packages/extension/src/ui/states/tokens.ts
janek26: agree, but I wanted to avoid introducing `immer` as a dependency

PR #187, opened 2022-01-21
feat: expose version to dapp

### LINKED ISSUE #183: Exposing the extension version to DApps
Expose the version of Argent X to DApps, so they can log it on error or display it, to the user, or whatever.

janek26:
closes #183

### DIFF
--- a/packages/extension/src/inpage/index.ts
+++ b/packages/extension/src/inpage/index.ts
@@ -12,7 +12,7 @@ import { MessageType, WindowMessageType } from "../shared/MessageType"
 import { getProvider } from "../shared/networks"
 import { EventHandler, StarknetWindowObject } from "./model"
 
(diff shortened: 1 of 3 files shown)


packages/extension/src/inpage/index.ts
```diff
@@ -12,6 +12,8 @@ import { MessageType, WindowMessageType } from "../shared/MessageType"
 import { getProvider } from "../shared/networks"
 import { EventHandler, StarknetWindowObject } from "./model"
 
+const VERSION = `v${process.env.VERSION}`
```
delaaxe, line 15: `v` should be only used for display purposes, not as part of the exposed value

delaaxe (changes requested):

delaaxe (approved):

PR #195, opened 2022-01-24
refactor(app): replace state machine by react router

delaaxe:
- Put all state in `zustand` hooks
- Route screen through `react-router` paths
- Remove `xstate`


packages/extension/src/ui/utils/transactions.ts
```diff
@@ -0,0 +1,38 @@
...
+import { sendMessage } from "../../shared/messages"
+
+interface TransactionRequest {
+  to: string
+  method: string
+  calldata: Args
+}
+
+export const sendTransaction = (
+  data: any /*TransactionRequest | InvokeFunctionTransaction*/,
```
janek26, line 13: avoid any

packages/extension/src/ui/utils/transactions.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import { sendMessage } from "../../shared/messages"
+
+interface TransactionRequest {
+  to: string
+  method: string
+  calldata: Args
+}
+
+export const sendTransaction = (
```
janek26, line 12: maybe we can just accept `InvokeFunctionTransaction`. If so lets remove this abstraction

janek26 (approved):
great work! lets make sure the global state gets less and less used with our future pr's and introduce more and more of those scoped states.

PR #200, opened 2022-01-25
chore: add eslint to webpack 

delaaxe:


packages/extension/src/background/actionQueue.ts
```diff
@@ -1,8 +1,8 @@
 import oHash from "object-hash"
-import { InvokeFunctionTransaction, typedData } from "starknet"
 
 import { getFromStorage, setToStorage } from "./storage"
 
+// eslint-disable-next-line @typescript-eslint/ban-types
 function objectHash(obj: {} | null) {
```
janek26, line 6:
```suggestion
function objectHash(obj: object | null) {
```

packages/extension/src/background/actionQueue.ts
```diff
@@ -1,8 +1,8 @@
 import oHash from "object-hash"
-import { InvokeFunctionTransaction, typedData } from "starknet"
 
 import { getFromStorage, setToStorage } from "./storage"
 
+// eslint-disable-next-line @typescript-eslint/ban-types
```
janek26, line 5:
```suggestion
```

packages/extension/src/background/index.ts
```diff
@@ -75,7 +75,7 @@ async function main() {
 
   messageStream.subscribe(async ([msg, sender]) => {
     const sendToTabAndUi = async (msg: MessageType) => {
-      sendMessageToActiveTabsAndUi(msg, [sender.tab?.id])
+      sendMessageToActiveTabsAndUi(msg)
```
janek26, line 78: this is needed to send messages to tabs that are **not**:
1. extension UI
2. connected dapps

This means it is needed to answer the sender swell if available, which would be the case for ie extension popups.

janek26 (changes requested):

packages/extension/src/background/index.ts
delaaxe: currently the argument is not used at all. let's add it back when it's actually needed? (no dead code, no "preparation" code on `main`)

janek26: my fault, it should be used. 
We should add 
```
await sendMessageToActiveTabs(message, additionalTargets)
```
in activeTabs.ts#49

packages/extension/src/background/activeTabs.ts
```diff
@@ -46,5 +46,5 @@ export async function sendMessageToActiveTabsAndUi(
   additionalTargets: Array<number | undefined> = [],
 ) {
   await sendMessageToUi(message)
-  await sendMessageToActiveTabs(message)
+  await sendMessageToActiveTabs(message, additionalTargets)
```
janek26, line 49: thank you 🙏

delaaxe: proof that this is useful :)

janek26 (approved):

PR #201, opened 2022-01-25
fix: transactions tracking

janek26:


packages/extension/src/background/trackTransactions.ts
```diff
@@ -87,7 +91,13 @@ export class TransactionTracker {
       }),
     )
     if (transactionStatus.length > 0) {
-      this.transactions = transactionStatus
+      // add transactions that were added while we were fetching
+      this.transactions = [
+        ...transactionStatus,
```
delaaxe, line 96: if array should be plural `transactionStatuses`

packages/extension/src/background/trackTransactions.ts
```diff
@@ -87,7 +91,13 @@ export class TransactionTracker {
...
+        ...this.transactions.filter((tx) => {
```
delaaxe, line 97: is it sure this concatenation will result in unique elements?

delaaxe: avoid abbreviations like `tx`

delaaxe (approved):

packages/extension/src/background/trackTransactions.ts
janek26: yes it does:
As a transaction hash is a unique identifier of a transaction and this operation is taking a superset of the fetched transactions and removes the intersection before merging

PR #204, opened 2022-01-26
fix(actions): restrict ActionScreen to specific routes 

delaaxe:


janek26:
why would you choose this over the approach we discussed? :-)
I think what we discussed is much more readable and easier to maintain

delaaxe:
> why would you choose this over the approach we discussed? :-) I think what we discussed is much more readable and easier to maintain

I tried but it didn't work. Also wasn't very readable / easy to understand what's actually going on

PR #205, opened 2022-01-26
fix(get-starknet): show popup ad when enabling without extension installed

delaaxe:


janek26:
Not sure this change solves the issue, does it? For me it looks the same, except that the argument is passed later. But the logic isn't touched, therefore it shouldn't change the behavior of the popup, does it?

**If we decide to keep this we would need to change the Readme.md accordingly**

delaaxe:
The logic is different because of the global state. I've updated the README

Readme.md
```diff
@@ -51,13 +47,13 @@ The package is a light wrapper around [starknet.js](https://github.com/seanjames
 ```javascript
 import { getStarknet } from "@argent/get-starknet"
 
-// check if wallet extension is installed and initialized. Shows a modal prompting the user to download ArgentX otherwise.
-const starknet = getStarknet({ showModal: true })
-const [userWalletContractAddress] = await starknet.enable() // may throws when no extension is detected
+const starknet = getStarknet() // May throw when no extension is detected
+// Check if wallet extension is installed and initialized. Shows a modal prompting the user to download Argent X otherwise.
```
janek26, line 51: This makes no sense to me :D `getStarknet()`may throws when no extension is detected? why? shouldn't `enable()` throw (and conditionally show the popup) if no extension is detected?

delaaxe: yeah was a bit too fast on this one, fixed

janek26 (approved):

PR #207, opened 2022-01-27
style: enforced curly braces to avoid sneaky errors

delaaxe:


packages/extension/src/background/activeTabs.ts
```diff
@@ -6,8 +6,12 @@ import { MessageType } from "../shared/MessageType"
...
 
 export function addTab(tabId?: number) {
-  if (tabId === undefined) return
-  if (hasTab(tabId)) return
+  if (tabId === undefined) {
+    return
+  }
+  if (hasTab(tabId)) {
+    return
+  }
```
janek26, line 14: combine with `||`

janek26 (approved):

packages/extension/src/background/activeTabs.ts
delaaxe: done

PR #208, opened 2022-01-27
feat: allow dapp to check if whitelisted

### LINKED ISSUE #202: Preauthorized sites
A common pattern with dapps is to have the wallet automatically connect when the user visits the website, but only if the user previously connected to it. This is to avoid firing the wallet connection window on first connect.

I propose the following:

 * Add a `isPreauthorized(): boolean` method to the signer interface. This method returns true if the current site is in the list of connected sites and calling `enable()` would not show a confirmation window to the user.

janek26:
closes #202

### DIFF
--- a/packages/get-starknet/src/extension.model.ts
+++ b/packages/get-starknet/src/extension.model.ts
@@ -23,6 +23,7 @@ type RpcMessage = {
 interface IStarketWindowObject {
   request: (call: RpcMessage) => Promise<void>
   enable: (options?: { showModal?: boolean }) => Promise<string[]>
+  isPreauthorized: () => Promise<boolean>
   on: (event: "accountsChanged", handleEvent: EventHandler) => void
   off: (event: "accountsChanged", handleEvent: EventHandler) => void
   signer?: SignerInterface
--- a/packages/get-starknet/src/index.ts
+++ b/packages/get-starknet/src/index.ts
@@ -6,26 +6,30 @@ import type { StarknetWindowObject } from "./extension.model"
 // nextjs ie needs this to be typeof window !== "undefined" as it's replacing it in client bundles
 const IS_BROWSER = typeof window !== "undefined"
 
+const fail = async () => {
+  throw Error("no starknet found in window")
+}
+
+const defaultStarknetWindowObject: StarknetWindowObject = {
+  request: fail,
+  isConnected: false,
+  provider: defaultProvider,
+  enable: ({ showModal } = {}) => {
+    if (IS_BROWSER && showModal) {
+      new App({ target: document.body })
+    }
+    return fail()
+  },
+  isPreauthorized: fail,
+  on: fail,
+  off: fail,
+  version: "uninstalled",
+}
+
 export const getStarknet = (): StarknetWindowObject => {
   // if extension isnt installed (didnt populate window.starknet) polyfill it
   if (!globalThis["starknet"]) {
-    const fail = async () => {
-      throw Error("no starknet found in window")
-    }
-    globalThis["starknet"] = {
-      request: fail,
-      isConnected: false,
-      provider: defaultProvider,
-      enable: ({ showModal } = {}) => {
-        if (IS_BROWSER && showModal) {
-          new App({ target: document.body })
-        }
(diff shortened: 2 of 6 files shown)


packages/get-starknet/src/extension.model.ts
```diff
@@ -23,6 +23,7 @@ type RpcMessage = {
 interface IStarketWindowObject {
   request: (call: RpcMessage) => Promise<void>
   enable: (options?: { showModal?: boolean }) => Promise<string[]>
+  isPreauthorized: () => Promise<boolean>
```
delaaxe, line 26: we should think of a way to avoid this duplication, eventually

packages/get-starknet/src/index.ts
```diff
@@ -6,26 +6,30 @@ import type { StarknetWindowObject } from "./extension.model"
...
-        if (IS_BROWSER && showModal) {
-          new App({ target: document.body })
-        }
-        return fail()
-      },
-      on: fail,
-      off: fail,
-      version: "uninstalled",
-    } as StarknetWindowObject
+    globalThis["starknet"] = defaultStarknetWindowObject
```
delaaxe, line 32: nice

packages/playground/src/pages/index.tsx
```diff
@@ -16,13 +17,20 @@ import styles from "../styles/Home.module.css"
 const Home: NextPage = () => {
   const [isConnected, setIsConnected] = useState(isWalletConnected())
   const [address, setAddress] = useState<string>()
+  const [preauthorized, setPreauthorized] = useState<boolean>(false)
```
delaaxe, line 20: Prefer starting boolean variables with `is` (as in `isPreauthorized`)

But more generally, a good codebase always uses the same terms to refer to a concept. Here `whitelist` and `preauthorization` seem to be the same thing, so we should choose a term and stick to it

Nb: same issue with `keystore` and `backup` but out of scope of this PR

packages/playground/src/pages/index.tsx
```diff
@@ -16,13 +17,20 @@ import styles from "../styles/Home.module.css"
...
   useEffect(() => {
     addWalletChangeListener((accounts) => {
       setAddress(accounts[0])
     })
   }, [])
 
+  useEffect(() => {
+    isPreauthorized().then((result) => {
+      setPreauthorized(result)
+    })
```
delaaxe, line 31:
```suggestion
    isPreauthorized().then(setPreauthorized)
```

?

packages/playground/src/pages/index.tsx
```diff
@@ -52,7 +60,12 @@ const Home: NextPage = () => {
             <button className={styles.connect} onClick={handleConnectClick}>
               Connect Wallet
             </button>
-            <p>First connect wallet to use dapp.</p>
+            <p>
+              First connect wallet to use dapp.
+              {preauthorized && (
+                <span> The application is already preauthorized</span>
+              )}
+            </p>
```
delaaxe, line 68: Should the dapp just auto connect if preauthorized instead of displaying this?

packages/playground/src/pages/index.tsx
janek26: I tend to agree. But I think preauthorize may be the better term tbh. It's just a whitelist of preauthorized websites. I'd like to change the implmentation to reflect that in a future PR and keep this as is for now as it's publicly facing.

packages/playground/src/pages/index.tsx
janek26: I thought having this is more explicit. 
The app's goal is to demo all features and how to use them. It doesn't need to implement the best UX possible, but poke details to devs.

packages/playground/src/pages/index.tsx
delaaxe: exactly, gotta be careful with public facing stuff because then it's much harder to change

if we're keeping this i'd suggest renaming all the "XXX_WHITELIST" event to "XXX_PREAUTHORIZATION" or something to keep things consistent and easy to lookup/understand

delaaxe (approved):

PR #222, opened 2022-01-28
feat(playground): autoconnect dapp if preauthorized

delaaxe:
All DeFi apps I use work this way, might as well do the same in our playground


janek26:
Like it 👍 but lets remove the useState then

janek26 (approved):

PR #257, opened 2022-02-08
feat(ui): delete localhost accounts

### LINKED ISSUE #232: Delete accounts
Localhost accounts can quickly clutter the ui and so should be deletable

delaaxe:
Allow deleting localhost accounts from the UI, even though they won't be persisted to the keystore.

Closes #232

### DIFF
--- a/packages/extension/src/ui/utils/messaging.ts
+++ b/packages/extension/src/ui/utils/messaging.ts
@@ -89,14 +89,14 @@ export const startSession = async (password: string): Promise<void> => {
 export const deleteAccount = async (address: string) => {
   sendMessage({ type: "DELETE_ACCOUNT", data: address })
 
-  const succeeded = await Promise.race([
-    waitForMessage("DELETE_ACCOUNT_RES").then(() => true),
-    waitForMessage("DELETE_ACCOUNT_REJ")
-      .then(() => false)
-      .catch(() => false),
-  ])
-
-  if (!succeeded) {
+  try {
+    await Promise.race([
+      waitForMessage("DELETE_ACCOUNT_RES"),
+      waitForMessage("DELETE_ACCOUNT_REJ").then(() => {
+        throw new Error("Rejected")
+      }),
+    ])
+  } catch (e) {
     throw Error("Could not delete account")
   }
 }


package.json
```diff
@@ -16,7 +16,7 @@
   "scripts": {
     "format": "prettier --loglevel warn --write \"**/*.{js,jsx,ts,tsx,css,md,yml,json}\"",
     "posttest": "npm run format",
-    "dev": "lerna run dev --parallel",
+    "dev": "lerna exec -- yarn dev",
```
janek26, line 19: what's the difference?

packages/extension/package.json
```diff
@@ -33,6 +33,7 @@
   },
   "dependencies": {
     "@extend-chrome/messages": "^1.2.2",
+    "@mui/icons-material": "^5.3.1",
```
janek26, line 36: have you talked to @graemeblackwood about this change? I guess the new icons won't match the old ones exactly

packages/extension/src/background/index.ts
```diff
@@ -464,7 +465,16 @@ async function main() {
         await downloadBackupFile()
         return sendToTabAndUi({ type: "DOWNLOAD_BACKUP_FILE_RES" })
       }
+
+      case "DELETE_ACCOUNT": {
+        try {
+          await deleteAccount(msg.data)
+          return sendToTabAndUi({ type: "DELETE_ACCOUNT_RES" })
+        } catch (error) {
+          console.error(error)
```
janek26, line 474: wanna keep it?

packages/extension/src/background/keys/l1.ts
```diff
@@ -12,10 +12,9 @@ const isDev = process.env.NODE_ENV === "development"
 
 interface StorageProps {
   encKeystore?: string
-  wallets: BackupWallet[]
 }
 
-const store = new Storage<StorageProps>({ wallets: [] }, "L1")
+const store = new Storage<StorageProps>({}, "L1")
```
janek26, line 17: like this! thanks

packages/extension/src/ui/utils/messaging.ts
```diff
@@ -85,3 +85,18 @@ export const startSession = async (password: string): Promise<void> => {
     throw Error("Wrong password")
   }
 }
+
+export const deleteAccount = async (address: string) => {
+  sendMessage({ type: "DELETE_ACCOUNT", data: address })
+
+  const succeeded = await Promise.race([
+    waitForMessage("DELETE_ACCOUNT_RES").then(() => true),
```
janek26, line 93: .catch(() => false)
to handle timeouts

package.json
delaaxe: I think it's more important to have colorized webpack output to more easily catch & parse errors. Starting time is the same

packages/extension/package.json
delaaxe: no but anyway it's better to use an icon library rather than fiddle with svg files. and the icon components can always be styled anyway

packages/extension/src/ui/utils/messaging.ts
delaaxe: fixed

graemeblackwood:
The icon change looks fine to me

delaaxe:
> The icon change looks fine to me

The back icon is slightly off but can always be fixed later

Issue #228, opened 2022-01-31
Allow User to remove tokens


janek26:
As a User I want to be able to remove/hide token balances of tokens that I don't use.

PR #259, opened 2022-02-09
feat: hide token

jgresham:
Implements #228 

https://user-images.githubusercontent.com/3721291/153275697-1b486a42-c696-4fcb-a352-9c0d271a6173.mp4

### DIFF
--- a/packages/extension/src/ui/screens/HideTokenScreen.tsx
+++ b/packages/extension/src/ui/screens/HideTokenScreen.tsx
@@ -1,0 +1,78 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
+import React, { FC, useState } from "react"
+import { useNavigate, useParams } from "react-router-dom"
+
+import { BackButton } from "../components/BackButton"
+import { Button, ButtonGroupVertical } from "../components/Button"
+import { Header } from "../components/Header"
+import { TokenIcon } from "../components/TokenIcon"
+import { P } from "../components/Typography"
+import { FormError, H2 } from "../components/Typography"
+import { routes } from "../routes"
+import { removeToken } from "../states/tokens"
+import { useTokensWithBalance } from "../states/tokens"
+import { toTokenView } from "../utils/tokens"
+import {
+  BalanceAlert,
+  TokenName,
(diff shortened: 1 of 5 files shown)


jgresham:
@janek26 let me know what you think

janek26:
I love it, great work! thanks so much.
We'll probably need to touch paddings and margins a bit in the warning screen, but that's a detail, thanks again!

packages/extension/src/ui/screens/HideTokenScreen.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export const HideTokenScreen: FC = () => {
+  const navigate = useNavigate()
+  const { tokenAddress } = useParams()
+  const { tokenDetails } = useTokensWithBalance()
+  const [error, setError] = useState("")
+
+  const token = tokenDetails.find(({ address }) => address === tokenAddress)
+  if (!token) {
+    return <></>
```
janek26, line 34: Maybe you can render a [`<Navigate />`](https://reactrouter.com/docs/en/v6/api#navigate) element here, to navigate back to the account page in case the token isn't found. I think that would be a little bit nicer error handling

janek26 (approved):
btw, appreciate the videos in your PRs

packages/extension/src/ui/screens/HideTokenScreen.tsx
jgresham: good idea, will make change tomorrow.

jgresham:
> I love it, great work! thanks so much. We'll probably need to touch paddings and margins a bit in the warning screen, but that's a detail, thanks again!

Sure, let the designers do their thing. Argent has great designers btw.

packages/extension/src/ui/screens/HideTokenScreen.tsx
jgresham: done

janek26 (approved):

delaaxe:
thanks @jgresham 
ok to release this as is for now, but as discussed with Janek, to be consistent with other confirmation screens, this should have been implemented in the `ActionScreen`, without an additional route

Issue #227, opened 2022-01-31
Large balance not shown correctly


janek26:
When the wallet amount is large enough, the page does not show the correct wallet balance
![unknown](https://user-images.githubusercontent.com/6831124/151778852-71e4233b-aa12-4f94-93ed-d9e65b030751.png)
![unknown21](https://user-images.githubusercontent.com/6831124/151778866-32d9a111-a0e9-4c1a-a962-2b7f7516a28b.png)

PR #260, opened 2022-02-09
fix: display large token balance & name in TokensList

jgresham:
Fixes: #227 

limit to 2 lines

### DIFF
--- a/packages/extension/src/ui/components/Token.tsx
+++ b/packages/extension/src/ui/components/Token.tsx
@@ -43,6 +43,15 @@ export const TokenTitle = styled.h3`
   line-height: 22px;
   margin: 0;
 `
+export const TokenMetaContainer = styled.div`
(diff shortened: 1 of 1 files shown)


janek26:
Thanks man :D 
I really like the handling of the name, balance could be a bit short imo, but we can improve on that later, maybe better to do it with ux folks.
We do have the same problem on the token details page, maybe you could add the same fixes to that page too (when you click a token you get to the detail page)

If you don't have time for that, that's also fine, let me know and I'll merge and create a follow up PR.

jgresham:
Thanks! 
The ellipsis take up a lot of space for balance.. but I suppose better than nothing for now.

Yeah, I'll do this for the token page tomorrow.

janek26 (approved):
As we're releasing today I want this to be in, centering can be done in a next version

PR #264, opened 2022-02-10
feat(ui): show better error message when deploying

delaaxe:
Avoid confusion about localhost network, with explanations about potential conflict on port 5000 on macOS.


janek26 (approved):
love it! good error messages!

Issue #263, opened 2022-02-10
Name Account


janek26:
As a user I want to be able to change the name of my account, to label them.

AC:
- Name changes do not need to be stored in backup files.
- I can change the displayed name of an account.

jgresham:
taking a look

jgresham:
Made some progress on this today.

1. Account names are saved in local storage using (networkId, address) as the key.
2. Changes the Profile icon using the account name
3. If the user tries to set a null account name, then "Account #" is used
4. Names are not saved in backup file


https://user-images.githubusercontent.com/3721291/153716498-1a36fa0d-2d1e-4da5-832b-54cd03e1aaa7.mp4

https://user-images.githubusercontent.com/3721291/153716500-4104c97d-978d-4c95-b4cc-d95caedc1520.mp4

juniset:
Hey @jgresham , that's great and exactly how it was intended to be implemented.

PR #284, opened 2022-02-12
feat: account names

jgresham:
Implements #263 
Needs cleanup once overall UI flow and functionality is approved.

https://user-images.githubusercontent.com/3721291/153716801-0204fcca-82fc-476d-944a-df65397bfc77.mp4

https://user-images.githubusercontent.com/3721291/153716798-e24eb527-78ca-49af-9ded-7eb830b58889.mp4


delaaxe:
Thanks for you contribution John, this is great! Small nitpick: it seems that when toggling edit mode, everything that's below the name shifts vertically by 1px. Possible to keep it all fixed?

Also, let me know when you think it's ready so I can review the code.

jgresham:
@delaaxe fixed the 1px UI shifting bug and cleaned up the PR.
Some things to consider:

1. EditableHeader isn't fully tested and may have repeat css from another component
2. account.ts might need some organization
3. in general, jsdoc strings would be helpful
4. wallet/account naming seems mixed #311 

note: this is separate from starknet's cli `--account=my_account`

packages/extension/src/ui/screens/ConfirmScreen.tsx
```diff
@@ -26,6 +26,7 @@ export interface ConfirmPageProps {
   selectedAccount?: {
     accountNumber: number
```
delaaxe, line 27: shouldn't be needed anymore

packages/extension/src/ui/states/account.ts
```diff
@@ -1,33 +1,120 @@
...
   selectedWallet,
-}: AccountStore) =>
-  Object.keys(wallets).findIndex((wallet) => wallet === selectedWallet) + 1
+}: AccountStore) => calcAccountNumber(wallets, selectedWallet)
+
+export const selectAccount = ({ wallets, selectedWallet }: AccountStore) => {
+  return selectedWallet ? wallets[selectedWallet] : undefined
+}
+
+const calcAccountNumber = (
```
delaaxe, line 113: please avoid abbreviations and acronyms as much as possible

packages/extension/src/ui/states/account.ts
```diff
@@ -1,33 +1,120 @@
...
+  wallets: Record<string, Wallet>,
+  accountAddress: string | undefined,
+): number => {
+  return (
+    Object.keys(wallets).findIndex((address) => address === accountAddress) + 1
+  )
+}
```
delaaxe, line 120: i think this should be more along the lines of:

```python
index = 1
while getDefaultAccountName(index) not in accountNames:
  index++
return getDefaultAccountName(index)
```

packages/extension/src/ui/utils/wallets.ts
```diff
@@ -38,14 +38,16 @@ export const getAccountColor = (accountNumber: number, withPrefix = true) =>
     argentColorsArray[(accountNumber % (argentColorsArray.length - 1)) + 1]
   }`
 
-export const getAccountImageUrl = (accountNumber: number) =>
-  `https://eu.ui-avatars.com/api?name=Account+${accountNumber}&background=${getAccountColor(
+export const getAccountImageUrl = (
+  accountName: string | undefined,
```
delaaxe, line 42:
```suggestion
  accountName?: string,
```

packages/extension/src/ui/utils/recovery.ts
```diff
@@ -34,6 +34,7 @@ export const recover = async ({
       .map(({ address, network }) => new Wallet(address, network))
       .reduce((acc, wallet) => ({ ...acc, [wallet.address]: wallet }), {})
 
+    setAccountNamesFromBackup(wallets)
```
delaaxe, line 37: this function wouldn't be needed if there was a single source of truth for account names

packages/extension/src/ui/Wallet.ts
```diff
@@ -11,13 +11,20 @@ const ArgentCompiledContractJson: CompiledContract = json.parse(
 export class Wallet {
   address: string
   networkId: string
+  name?: string
```
delaaxe, line 14: this field shouldn't exist on this class

delaaxe (changes requested):

packages/extension/src/ui/states/account.ts
```diff
@@ -1,33 +1,120 @@
...
-export const useAccount = create<AccountStore>((set) => ({
-  wallets: {},
-  addWallet: (newWallet: Wallet) =>
-    set((state) => ({
-      selectedWallet: newWallet.address,
-      wallets: {
-        ...state.wallets,
-        [newWallet.address]: newWallet,
+export const useAccount = create<AccountStore>(
+  persist(
```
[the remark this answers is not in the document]
jgresham: I think it could be cleaner and easier to reason about where the data is coming from. I kept in account store because I thought it would be less changes elsewhere. Few thoughts:

Pros for separate stores:
1. Keep the data coming from `recovery.ts recover()` (account store before these changes) separate from any user input/metadata
2. More future proof if there are plans for user's to add more account details.

Cons for separate stores:
1. UI components would need to use multiple stores to show all the data for an account

packages/extension/src/ui/states/account.ts
jgresham: I don't think this would work with the structure of `accountNames` being:
https://github.com/argentlabs/argent-x/pull/284/files#diff-7b1dbae5fd43e7fa9c15da5204f37abd705b208adc0be83178e26bcd3816c32dR48-R54
```
accountNames: {
              ...state.accountNames,
              [networkId]: {
                ...state.accountNames[networkId],
                [address]: name,
              },
            },
```

packages/extension/src/ui/utils/recovery.ts
jgresham: correct, but would need to add an additional "useStore" to UI components.

packages/extension/src/ui/Wallet.ts
jgresham: I'm not aware how `Wallet`  and `Account` is used with starknet. Is this class better named `Account`?

delaaxe:
@jgresham I've made the changes. Do you see anything that could be wrong or that I might've missed before I merge? Thanks again for your help

packages/extension/src/ui/states/accountMetadata.ts
```diff
@@ -0,0 +1,63 @@
...
+    useAccountMetadata.setState({ accountNames: names })
+  }
+}
+
+const getDefaultAccountName = (account: Wallet) => {
+  const names = Object.values(
+    useAccountMetadata.getState().accountNames[account.networkId] || {},
+  )
+  let index = 1
+  while (names.includes(`Account ${index}`)) {
```
jgresham, line 59: I think this method of determining the default account name has a bug. On first open of my wallet, all accounts had "Account 1" as the name. Also, my opinion is that the n-th account should be named Account n, regardless of the other account names.
account-numbers

jgresham: Account names may need to be set to the default name (if not set) after `recovery` to fix this.

packages/extension/src/ui/states/accountMetadata.ts
```diff
@@ -0,0 +1,63 @@
...
+  accountNames: Record<string, Record<string, string>>,
+): string => accountNames[networkId]?.[address] || "Unnamed account"
+
+export const setDefaultAccountNames = (accounts: Record<string, Wallet>) => {
+  const { accountNames } = useAccountMetadata.getState()
+  let names = accountNames
+  for (const [address, account] of Object.entries(accounts)) {
+    const { networkId } = account
+    if (!accountNames?.[networkId]?.[address]) {
+      const name = getDefaultAccountName(account)
```
jgresham, line 42: `names` may need to be passed in here as the state isn't updated for each iteration in `getDefaultAccountName`

delaaxe:
@jgresham just reverted it to the way you did it before. seeing any more issues?

jgresham:
> @jgresham just reverted it to the way you did it before. seeing any more issues?

I don't see any new commits?

delaaxe:
> > @jgresham just reverted it to the way you did it before. seeing any more issues?
> 
> I don't see any new commits?

sorry I pushed to wrong branch. should be ok now

jgresham:
@delaaxe just tested it out. looks good to me!

delaaxe (approved):

PR #323, opened 2022-02-20
fix: setIsDeployed true after account deployed

### LINKED ISSUE #322: "Undeployed" after an account is deployed
undeployedBug2

jgresham:
closes #322


delaaxe (changes requested):
great! thanks for this

delaaxe:
it seems I can't reproduce the original problem even on the main branch?

jgresham:
> it seems I can't reproduce the original problem even on the main branch?

I just verified that it occurs on main branch. Did you keep the extension open while it was deploying? If you close the extension and re-open, the "Undeployed" error goes away.

PR #326, opened 2022-02-21
feat: different signing key for each account

gergold:
A new object `Wallet` is instantiated in the background script and is replacing all the logic contained in `l1.ts` file.
The `Wallet` object holds accounts (`WalletAccount`) and is responsible for:
* private keys management (a different key is now derived for each account)
* all operations on account (add, remove, select)
* persistence in local storage

⚠️ This is not backward compatible with current version, the local storage key is different and the backup file is not compatible. This is supposed to be released with the Proxy account which is not backward compatible too.

Unit tests have been added to test the key derivation mechanism and the wallet features


packages/extension/src/background/wallet.ts
```diff
@@ -0,0 +1,275 @@
...
+    const url = URL.createObjectURL(blob)
+    const filename = "argent-x-backup.json"
+    return { url, filename }
+  }
+
+  public static validateBackup(backupString: string): boolean {
+    try {
+      const backup = JSON.parse(backupString)
+      return ajv.validate(backupSchema, backup)
+    } catch (error) {
```
delaaxe, line 227: can remove error variable if unused

```suggestion
    } catch {
```

delaaxe (commented):
this is great, thank you, I really like this new Wallet class that makes the background code a lot less spaghetti-ish.

Issue #94, opened 2021-12-19
Can't recover wallet on Linux


parseb:
I might be totally wrong as this can be a particularity of my setup. 

Can anyone actually drop a file in that space?
In order to drag _starknet-backup.json_ you likely need to click somewhere else, which closes the dialogue and makes restarting the process a necessity. The process itself (browsing, selecting the backup file) does nothing for me, but not working is less annoying than "promising" me a drag-and-drop experience and not delivering. (test environment: Ubuntu / Chromium / v1.1.0) 

![image](https://user-images.githubusercontent.com/5999852/146692110-93e51e09-7881-48be-bee8-b91d0ea6c62c.png)

arya4291:
I can confirm I have the same issues using Ubuntu & Brave

aramanja:
I can confirm I have the same issues when using Ubuntu & Brave. The drag & drop function and json restoration works fine for me on Windows 10 & Brave.

janek26:
it seems like drag and drop is an issue for ubuntu, please let me know if you have any way of fixing this.
Restoring after that should work just fine.

parseb:
The drag&drop functionality is OS correlated behavior. It also happens for metamask for example. 
I do not think that in itself is a bug. 

Suggested First step:
- Reproduce wallet recovery dis-functionality on ubuntu/brave or ubuntu/chromium.

jgresham:
Looks like the most recent bug report to chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=1027558
Haven't found one for linux/OS

janek26:
Hey @jgresham, thanks for looking into this. We already thought of a simple solution, basically opening the setup process in a dedicated tab, in case we detect ubuntu/linux.
This could also transform into the default behavior when we decide to support ledger, they need a dedicated tab anyways to connect to their devices.

janek26:
I can publish more around this soon, and create a few new issues to the project, but it's already late here :D

jgresham:
Found a workaround here that works now!

1. Open extension
2. Right click "Inspect" (Chrome devtools window should open)
3. Import file now and chrome should not close extension when you focus on file selector

gergold:
Another workaround is to open the extension in a tab using its url:
`chrome-extension://dlcobpjiigpikoobohmabehhmhfoodbb/index.html`
and navigate to "Restore Wallet".

Anyway, we are working on a button in the Settings that will open the extension in a new tab

PR #329, opened 2022-02-23
feat: open recover screen in a tab on linux

gergold:
This is fixing issue #94 
On linux, when pressing "Recover Wallet" it will open the extension in a new browser tab.

It's not opening the recover screen but the welcome screen as in order to open that app at a given screen we need to change the main router

### DIFF
--- a/packages/extension/src/ui/hooks/useCustomNavigate.ts
+++ b/packages/extension/src/ui/hooks/useCustomNavigate.ts
@@ -1,0 +1,30 @@
+import { NavigateOptions, To, useNavigate } from "react-router-dom"
+import browser from "webextension-polyfill"
+
+const openExtensionInTab = () => {
+  const url = browser.runtime.getURL(`index.html`)
+  return browser.tabs.create({ url })
+}
+
+const isAlreadyInTab = async () => {
+  return Boolean(await browser.tabs.getCurrent())
+}
+
+const getPlatformOS = async () => {
+  const info = await browser.runtime.getPlatformInfo()
+  return info.os
+}
+
+export const useCustomNavigate = () => {
+  const navigate = useNavigate()
+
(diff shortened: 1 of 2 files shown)


packages/extension/src/ui/hooks/useCustomNavigate.ts
```diff
@@ -0,0 +1,30 @@
+import { NavigateOptions, To, useNavigate } from "react-router-dom"
+import browser from "webextension-polyfill"
+
+const openExtensionInTab = () => {
+  const url = browser.runtime.getURL(`index.html`)
```
delaaxe, line 5:
```suggestion
  const url = browser.runtime.getURL("index.html")
```

packages/extension/src/ui/hooks/useCustomNavigate.ts
```diff
@@ -0,0 +1,30 @@
...
+const isAlreadyInTab = async () => {
+  return Boolean(await browser.tabs.getCurrent())
+}
+
+const getPlatformOS = async () => {
+  const info = await browser.runtime.getPlatformInfo()
+  return info.os
+}
+
+export const useCustomNavigate = () => {
```
delaaxe, line 18: if the goal is really just to circumvent Linux limitations, why not name it `useLinuxNavigate` to be more explicit about where it's needed rather than the pretty generic `custom`?

delaaxe (approved):

packages/extension/src/ui/hooks/useCustomNavigate.ts
gergold: The idea is to use it for future needs, for example, when connecting a Ledger, the extension needs to be in a tab in order to allow chrome to access USB device.

PR #345, opened 2022-02-26
fix(dapp): unhandled error when adding token in Dapp

jasonzhouu:


packages/dapp/src/components/TokenDapp.tsx
```diff
@@ -25,6 +25,9 @@ export const TokenDapp: FC = () => {
   const [transactionStatus, setTransactionStatus] = useState<
     "idle" | "approve" | "pending" | "success"
   >("idle")
+  const [errorMessageOfAddERC20Token, setErrorMessageOfAddERC20Token] =
```
delaaxe, line 28:
```suggestion
  const [addTokenError, setAddTokenError] =
```

packages/dapp/src/components/TokenDapp.tsx
```diff
@@ -25,6 +25,9 @@ export const TokenDapp: FC = () => {
...
+    useState("")
+  const [errorMessageOfAddETHToken, setErrorMessageOfAddETHToken] = useState("")
```
delaaxe, line 30: One error message variable should be enough

packages/dapp/src/components/TokenDapp.tsx
```diff
@@ -207,7 +210,15 @@ export const TokenDapp: FC = () => {
...
-          onClick={() => addToken(tokenAddress)}
+          onClick={() => {
+            addToken(tokenAddress)
+              .then(() => {
+                setErrorMessageOfAddERC20Token("")
+              })
+              .catch((error) => {
+                setErrorMessageOfAddERC20Token(error.message)
+              })
+          }}
```
delaaxe, line 221: Please prefer async/await notation

packages/dapp/src/components/TokenDapp.tsx
```diff
@@ -221,14 +232,23 @@ export const TokenDapp: FC = () => {
...
             style={{ marginLeft: ".6em" }}
-            onClick={() => addToken(ethAddress)}
+            onClick={() => {
+              addToken(ethAddress)
+                .then(() => {
+                  setErrorMessageOfAddETHToken("")
+                })
+                .catch((error) => {
+                  setErrorMessageOfAddETHToken(error.message)
+                })
```
delaaxe, line 250: Please prefer async/await notation

delaaxe (changes requested):

packages/dapp/src/components/TokenDapp.tsx
jasonzhouu: If only one variable is used, an error occurs when adding the erc20 token, the error message will be displayed in both places

packages/dapp/src/components/TokenDapp.tsx
jasonzhouu: OK. updated

packages/dapp/src/components/TokenDapp.tsx
jasonzhouu: updated

packages/dapp/src/components/TokenDapp.tsx
delaaxe: It's fine displaying just one error in one place

delaaxe (approved):

jasonzhouu:
Thank you

Issue #293, opened 2022-02-14
Improve UX for hiding tokens


delaaxe:
To be consistent across the UI, the `HideTokenScreen` should reuse the existing `ConfirmScreen` (without going through the Actions flow)

janek26:
closed by #346

PR #346, opened 2022-02-26
refactor(ui): reuse ConfirmScreen in HideTokenScreen

jasonzhouu:
Implements https://github.com/argentlabs/argent-x/issues/293

https://user-images.githubusercontent.com/19926174/155840471-eb232766-6c8c-4ac8-a3c3-2aab9ed02d32.mov

### DIFF
--- a/packages/extension/src/ui/screens/HideTokenScreen.tsx
+++ b/packages/extension/src/ui/screens/HideTokenScreen.tsx
@@ -26,11 +26,11 @@ export const HideTokenScreen: FC = () => {
 
   const { name } = toTokenView(token)
 
-  const handleSubmit = async () => {
+  const handleSubmit = () => {
     try {
       removeToken(token)
       navigate(routes.account())
-    } catch (e) {
+    } catch {
       setError("Token not hidden")
     }
   }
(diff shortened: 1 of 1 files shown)


packages/extension/src/ui/screens/HideTokenScreen.tsx
```diff
@@ -35,43 +26,33 @@ export const HideTokenScreen: FC = () => {
 
   const { name } = toTokenView(token)
 
-  return (
-    <>
-      <Header>
-        <BackButton />
-      </Header>
+  const handleSubmit = async () => {
```
delaaxe, line 29: Does this function need to be async?

packages/extension/src/ui/screens/HideTokenScreen.tsx
```diff
@@ -35,43 +26,33 @@ export const HideTokenScreen: FC = () => {
...
+    try {
+      removeToken(token)
+      navigate(routes.account())
+    } catch (e) {
```
delaaxe, line 33: Hide error variable if unused
```suggestion
    } catch {
```

delaaxe (changes requested):
This is much cleaner, thank you

packages/extension/src/ui/screens/HideTokenScreen.tsx
jasonzhouu: Thank you for your review. 
Updated

delaaxe (approved):

Issue #292, opened 2022-02-14
Emit event when a dapp is disconnected


juniset:
When the user disconnects a dapp, the extension should pass an event to the dapp.

We should use the `accountsChanged` event but pass an empty array, see https://github.com/argentlabs/argent-x/blob/df7c174d26c435c6d951d6d9d4fd5ebbd082d761/packages/extension/src/inpage/index.ts#L132

See #262 for more context.

jgresham:
It doesn't look like the metamask api docs support what we were hoping to do. We would need to pass the `accountsChanged` (with empty array) event to only the dapp that we are disconnecting from. Also, for argent-x I don't see the ability to pass a message from the extension to a specific dapp/tab/host. Am I missing anything?

jasonzhouu:
actually Metamask do send message from extension to specific tab

https://github.com/MetaMask/metamask-extension/blob/19098f7d69c5f1ca7f2cfd31f7179e0a44d7db2f/app/scripts/metamask-controller.js#L3486

juniset:
Closed by #378

PR #351, opened 2022-02-27
feature: emit event when a dapp is disconnected

jasonzhouu:
Implements https://github.com/argentlabs/argent-x/issues/292

emit `accountsChanged` event with data of empyt array

https://user-images.githubusercontent.com/19926174/155878786-121326d7-d7f5-4e73-bc7d-e61d31af0a1b.mov


jasonzhouu:
`activeTabs` is change from set of tab id to array of object like:


```
[{tabId: 526, host: 'localhost:3000'}]
```


After disconnect from extension, tabIds with corresponding host will be found from here.

gergold:
This should be covered by #378 
So closing this PR

PR #375, opened 2022-03-08
Localhost: always deploy ArgentAccount

gergold:


packages/extension/src/background/index.ts
```diff
@@ -1,3 +1,4 @@
+import ArgentAccountCompiledContract from "!!raw-loader!../contracts/ArgentAccount.txt"
```
janek26, line 1: 🤮
hopefully starknet will fix this soon

packages/extension/src/background/wallet.ts
```diff
@@ -154,11 +156,27 @@ export class Wallet {
     this.accounts.push(account)
 
     await this.writeBackup()
-    await this.selectAccount(deployTransaction.address)
+    await this.selectAccount(account.address)
 
     return { account, txHash: initTransaction.transaction_hash }
```
janek26, line 161: does it show as 2 transactions in the ui? or are we just dropping the implementation transaction from the transaction tracker?

janek26 (approved):
love the tests. Few nitpicks, but ok overall

packages/extension/src/background/wallet.ts
gergold: yes, we just track the second one... until Starknet implements the factory pattern and we can bundle them into a single tx

PR #377, opened 2022-03-08
feat: add Ledger account and blind signing

jcaporossi-ledger:
Hello,

 Integration of Ledger Nano device (blind signing only).


gergold:
Agree with @janek26 
The branch spike/ledger-support was written before the starknet v3 changes, this is why the abstraction was done in ArgentX. Now that we have a Signer class in starknet.js it will be much more easier.

jcaporossi-ledger:
Ok I will have a look into starknet Signer class, it makes sense yes...

gergold:
If you look into that, I think it would make a lot of sense to publish it as a new npm package (something like `starknet-ledger-signer`) that can be imported by any project.

jcaporossi-ledger:
Ok it is noted, thanks for your feedback

jcaporossi-ledger:
Hi, 

 Just a question with npm package as I am not quite familiar with them. I have made modifications in starknet.js locally (in a dedicated branch), build and tests are OK. Then I would like to test this package with argent-x extension. In my starknet.js local repo, I have performed a `npm link`. Then switch into argent-x/packages/extension/,  `npm link starknet` in order to compile argent-x with my local starknet package. However, after trying to build the extension with `yarn build`, I have many errors with starknet. What did I do wrong ?

jcaporossi-ledger:
Many errors due to webpack as this message is always displayed :

```
ERROR in ../../../starknet.js/dist/utils/encode.js 37:63-69
Module not found: Error: Can't resolve 'buffer' in '/Users/jerome.caporossi/Documents/starknet.js/dist/utils'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
	- add a fallback 'resolve.fallback: { "buffer": require.resolve("buffer/") }'
	- install 'buffer'
If you don't want to include a polyfill, you can use an empty module like this:
	resolve.fallback: { "buffer": false }
 @ ../../../starknet.js/dist/index.js 38:30-55
 @ ./src/inpage/index.ts 78:17-36
 
```

jcaporossi-ledger:
Ok found what's going wrong: had to install "buffer" package in my local starknet package....

jcaporossi-ledger:
Guys, just notice that currently argent-x extension relies on starknet.js v2, and starknet v3 introduces many modifications regarding Signer interface as Signer does not inherit from Provider as in v2....

dhruvkelawala:
> Guys, just notice that currently argent-x extension relies on starknet.js v2, and starknet v3 introduces many modifications regarding Signer interface as Signer does not inherit from Provider as in v2....

That's why Argent-X is also getting updated to v3 soon to make use of AccountInterface of starknet v3

janek26:
@dhruvkelawala is right.
please work with starknet.js develop branch (v3) and argent x develop branch (also ready for v3)

This PR gets closed in favor of https://github.com/0xs34n/starknet.js/pull/139

Issue #358, opened 2022-03-03
update `isPreauthorized()` method to check if the wallet is signed out


dhruvkelawala:
- Check if wallet is signed out before checking if the site is whitelisted
- If the Wallet is signed out, prompt user to sign in like MetaMask.
- After successful sign in, check the dapp in whitelist

juniset:
Closed by #378

PR #348, opened 2022-02-26
feature: disconnect from dapp

jasonzhouu:
Implements  https://github.com/argentlabs/argent-x/issues/262

https://user-images.githubusercontent.com/19926174/155849631-67924e4e-4618-4153-a907-f38f0981a87d.mov

https://user-images.githubusercontent.com/19926174/155849796-7e4bf8e1-819d-4914-a357-95b0a5d66652.mov

## PS
npm package `@argent/get-starknet` should be upgraded in the Dapp.

Issue #295, opened 2022-02-15
Accounts changed event is not fired


fracek:
I noticed that even if my application subscribes to the `accountsChanged` event (using `starknet.on('accountsChanged', cb)`, my callback is never called.

jgresham:
@fracek there is an example dapp with this functionality working. Here is the example code: 
https://github.com/argentlabs/argent-x/blob/c7540a82dcbd34776a6a1bad79d2451f0c26482a/packages/dapp/src/services/wallet.service.ts#L97

juniset:
Closed by #378

PR #378, opened 2022-03-08
Feature/unauthorize dapp

gergold:
Those two actions: 

* un-authorizing a dapp in the settings, 
* locking the wallet,

will clear the account and address from the injected `starknet` object and the dapp can adapt accordingly (disconnect). The code is inspired from #348 and #351 from @jasonzhouu which were not merges as they are since `develop` branch had evolved a lot since then.

We **don't implement** a `disconnect()` method that would allow a dapp to remove it's own authorization because we believe that should happen in the extension.

This PR should solve these issues too:
* #292
* #295
* #358

### DIFF
--- a/packages/extension/src/background/activeTabs.ts
+++ b/packages/extension/src/background/activeTabs.ts
@@ -3,11 +3,21 @@ import browser from "webextension-polyfill"
 import { sendMessage } from "../shared/messages"
 import { MessageType } from "../shared/MessageType"
 
-const activeTabs = new Set<number>()
+type Tab = { id: number; host: string }
+const activeTabs: Tab[] = []
 
-export function addTab(tabId?: number) {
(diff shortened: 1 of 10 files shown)


packages/extension/src/background/activeTabs.ts
```diff
@@ -3,32 +3,65 @@ import browser from "webextension-polyfill"
...
 
 export function hasTab(tabId?: number) {
   if (tabId === undefined) {
     return false
   }
-  return activeTabs.has(tabId)
+  return activeTabs.some((tab) => tab.id === tabId)
 }
 
 browser.tabs.onRemoved.addListener((tabId) => {
```
delaaxe, line 31: Would:

```suggestion
browser.tabs.onRemoved.addListener(removeTab)
```

be more graceful?

delaaxe (approved):

PR #379, opened 2022-03-08
feat(ui): onboarding screen

delaaxe:
- better ux for onboarding screen
- add option to not be asked again about backup download


janek26 (approved):
code looks good! built and tested

PR #383, opened 2022-03-09
feat(extension): support legacy transactions

delaaxe:
Support `starknet.js` v2 code in dapps for backwards compatibility and a smoother migration

### DIFF
--- a/packages/extension/src/inpage/index.ts
+++ b/packages/extension/src/inpage/index.ts
@@ -13,6 +13,7 @@ import {
 
 import { MessageType, WindowMessageType } from "../shared/MessageType"
 import { getProvider } from "../shared/networks"
+import { LEGACY_WalletSigner } from "./legacy"
 import { EventHandler, StarknetWindowObject } from "./model"
 
 const VERSION = `${process.env.VERSION}`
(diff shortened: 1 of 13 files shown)


packages/extension/src/inpage/index.ts
```diff
@@ -119,6 +120,7 @@ const starknetWindowObject: StarknetWindowObject = {
           const { address, network } = data.data
           starknet.provider = getProvider(network)
           starknet.account = new ArgentXAccount(address, starknet.provider)
+          starknet.signer = new LEGACY_WalletSigner(address, starknet.provider)
```
gergold, line 123: Make sure to merge `develop`or rebase, and set `starknet.signer = undefined` when `DISCONNECT_ACCOUNT` is emmited

packages/extension/src/inpage/model.ts
```diff
@@ -34,6 +38,7 @@ interface IStarketWindowObject {
 
 interface ConnectedStarketWindowObject extends IStarketWindowObject {
   isConnected: true
+  signer: SignerInterface
```
gergold, line 41: This is the `SignerInterface` of v3, is that correct then?

gergold (approved):

packages/extension/src/inpage/index.ts
delaaxe: fixed

packages/extension/src/inpage/model.ts
delaaxe: It's kind of a mix of the two (because of laziness)... The most important is that the `invokeFunction`'s prototype is the one from v2, I guess I should check if the main dapps use any of the other methods

PR #386, opened 2022-03-10
feat(ui): add edit icon next to account title

delaaxe:
- add edit icon when hovering account title
- fix empty title bug
- finish editing by pressing enter


janek26 (approved):
tested

PR #387, opened 2022-03-10
feat(artifacts): remove debug_info key and minify json

gergold:
remove `program.debug_info` key and minify json
```
ArgentAccount.txt: 5.2MB  => 556KB
Proxy.txt:         618 KB => 91KB
```


gergold:
`program.debug_info` is maybe used when a tx fails to provide error trace, so keeping it for now!
Just minify json then:
```
ArgentAccount.txt: 5.2 MB  => 2.0 MB
Proxy.txt:         618 KB => 258 KB
```

PR #391, opened 2022-03-10
fix: token details fetch

janek26:
this was caused by the starknet v3 update.
It's very likely that we need to touch this again once contracts interface v3 is merged


delaaxe (approved):

packages/extension/src/ui/screens/AddTokenScreen.tsx
```diff
@@ -132,7 +132,7 @@ export const AddTokenScreen: FC<AddTokenScreenProps> = ({
       </Header>
 
       <AddTokenScreenWrapper>
-        <H2>Add token</H2>
+        <H2>Add tokens</H2>
```
delaaxe, line 135: Not sure about the plural here

PR #420, opened 2022-03-25
Spike/ledger support the revenge

ghost:


yogh333:
Hello,
Sorry but I have had to delete my github account (jcaporossi-ledger) the PR was made from. The repo has been transferred to the new github account @yogh333.

janek26:
hey! may we use webusb as well? like tally does here: https://github.com/tallycash/extension/blob/main/background/services/ledger/index.ts

We also discussed ledger inside of starknet.js and I think we'll move it into its own package, as it introduces new dependencies like buffer and all the ledger deps

janek26:
closed by #694

Issue #444, opened 2022-03-31
NFTs balance uses 18 decimals rather than no decimals


ametel01:
after adding an ERC721 to the tokens and receiving 1 NFT the balance will show 0.000000000000000001 rather than 1.

janek26:
We're assuming ERC20 tokens here, and 18 is a good default for that.
ERC721 does not expose a `decimals` method unchain, so we'll look into it if we can have some smart ERC721 detection

PR #445, opened 2022-03-31
changed default number of decimals to 0 #444

ametel01:


janek26:
https://github.com/argentlabs/argent-x/issues/444#issuecomment-1085547977

PR #446, opened 2022-04-01
feat: add PENDING state as final

janek26:
As StarkWare confirmed the PENDING state used by StarkNet is as final as ACCEPTED_ON_L2, therefore we should treat it the same to speed up (mainnet) transactions

### DIFF
--- a/packages/extension/src/background/index.ts
+++ b/packages/extension/src/background/index.ts
@@ -60,14 +60,14 @@ const successStatuses = ["ACCEPTED_ON_L1", "ACCEPTED_ON_L2", "PENDING"]
       for (const transaction of transactions) {
         const { hash, status, accountAddress, meta } = transaction
         if (
-          (successStatus.includes(status) || status === "REJECTED") &&
+          (successStatuses.includes(status) || status === "REJECTED") &&
           !(await hasShownNotification(hash))
         ) {
           addToAlreadyShown(hash)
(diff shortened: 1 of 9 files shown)


packages/extension/src/background/index.ts
```diff
@@ -37,6 +37,8 @@ import { Storage, clearStorage } from "./storage"
 import { TransactionTracker, getTransactionStatus } from "./trackTransactions"
 import { Wallet, WalletStorageProps } from "./wallet"
 
+const successStatus = ["ACCEPTED_ON_L1", "ACCEPTED_ON_L2", "PENDING"]
```
delaaxe, line 40:
```suggestion
const successStatuses = ["ACCEPTED_ON_L1", "ACCEPTED_ON_L2", "PENDING"]
```

packages/extension/webpack.config.js
```diff
@@ -69,6 +69,7 @@ module.exports = {
   ],
   resolve: {
     extensions: [".tsx", ".ts", ".js"],
+    fallback: { buffer: require.resolve("buffer/") },
```
delaaxe, line 72: already in previous PR

delaaxe (approved):

packages/extension/webpack.config.js
janek26: should auto resolve

PR #475, opened 2022-04-07
feat: activity screen

delaaxe:


packages/extension/src/ui/components/Account/AccountActivity.tsx
```diff
@@ -14,19 +20,44 @@ const Header = styled.h2`
...
+  const { switcherNetworkId } = useAppState()
+  const [activity, setActivity] = useState<DailyActivity>()
+
+  useEffect(() => {
+    ;(async () => {
+      if (!activity) {
+        setActivity(await fetchActivity(account.address, switcherNetworkId))
+      }
+    })()
+  }, [activity])
```
janek26, line 43: https://swr.vercel.app

janek26:
voyager said they're ready to take the load

packages/extension/src/ui/components/Account/AccountActivity.tsx
```diff
@@ -0,0 +1,63 @@
...
+  account: Account
+}
+
+export const AccountActivity: FC<AccountActivityProps> = ({ account }) => {
+  const { switcherNetworkId } = useAppState()
+
+  const { data: activity = {} } = useSWR(
+    [account.address, switcherNetworkId],
+    fetchActivity,
+    { suspense: false },
```
janek26, line 39: 1. should we use suspense to show a loading spinner in case there's no stale data? :) should be the same as in assets
2. a [ErrorBoundary](https://reactjs.org/docs/error-boundaries.html) ontop to be prepared in case the api fails? (to keep the api dependencies low)
3. [add interval](https://swr.vercel.app/docs/revalidation#revalidate-on-interval) (60s) to always have fresh data?

delaaxe: 1. since it's loading values from cache the spinner would be visible only once
2. maybe
3. yes

delaaxe: "Only class components can be error boundaries."

packages/extension/src/ui/components/Account/AccountActivity.tsx
```diff
@@ -34,9 +34,9 @@ export const AccountActivity: FC<AccountActivityProps> = ({ account }) => {
   const { switcherNetworkId } = useAppState()
 
   const { data: activity = {} } = useSWR(
-    [account.address, switcherNetworkId],
+    [account.address, switcherNetworkId, "activity"],
     fetchActivity,
-    { suspense: false },
+    { refreshInterval: 60e3 /* 1 minute */ },
```
janek26, line 39: love this

PR #476, opened 2022-04-08
Feature/replace-backup-file-by-seed-phrase

janek26:
https://user-images.githubusercontent.com/6831124/162443094-b0ff288d-5821-4b85-97af-dee23b1969b6.mp4


packages/extension/src/background/wallet.ts
```diff
@@ -64,6 +64,18 @@ export class Wallet extends EventEmitter {
     return this.session !== undefined
   }
 
+  public async getSeedPhrase(): Promise<string> {
+    if (!this.isSessionOpen() || !this.session || !this.encryptedBackup) {
```
delaaxe, line 68: Looks like we're not so confident on the invariants this class is supposed to hold

packages/extension/src/ui/App.tsx
```diff
@@ -83,14 +85,10 @@ const Screen: FC = () => {
...
   const { isLoading } = useAppState()
   const { actions } = useActions()
-  const { isBackupDownloadRequired } = useBackupDownload()
 
   if (isLoading) {
     return <LoadingScreen />
   }
-  if (isBackupDownloadRequired) {
-    return <BackupDownloadScreen />
-  }
```
delaaxe, line 93: so no more backup file at all?

packages/extension/src/ui/App.tsx
```diff
@@ -114,6 +112,18 @@ const Screen: FC = () => {
...
+            element={<ConfirmSeedRecoveryPage />}
+          />
+          <Route
+            path={routes.setupSeedRecovery()}
+            element={<SetupSeedRecoveryPage />}
+          />
+          <Route
+            path={routes.setupRecovery()}
+            element={<SetupRecoveryPage />}
+          />
```
delaaxe, line 126: we should increase max line width

packages/extension/src/ui/components/Account/UpdateBanner.tsx
```diff
@@ -1,70 +1,38 @@
...
-        <UpdateBannerDescription>
-          This may have breaking changes
-        </UpdateBannerDescription>
-      </UpdateTextWrapper>
-    </UpdateBannerWrapper>
+    <Banner
+      title="Update Available"
+      description="This may have breaking changes"
+      icon={
+        <svg
```
delaaxe, line 15: Should be in its own file

packages/extension/src/ui/components/Banner.tsx
```diff
@@ -0,0 +1,65 @@
...
+  cursor: pointer;
+  align-items: center;
+  padding: 16px;
+  background-color: #ffffff;
+  border-radius: 8px;
+  ${({ noMargins = false }) =>
+    !noMargins &&
+    css`
+      margin: 16px 20px;
+    `};
```
delaaxe, line 17: `styled-components` is really an awful css library

packages/extension/src/ui/components/FeeEstimate.tsx
```diff
@@ -3,7 +3,6 @@ import DangerIcon from "@mui/icons-material/ReportGmailerrorredRounded"
 import Tippy from "@tippyjs/react"
```
delaaxe, line 3: component and filename should either be called `FeeEstimation` or `EstimateFee`

packages/extension/src/ui/components/Recovery/IconBar.tsx
```diff
@@ -0,0 +1,92 @@
...
+  align-items: center;
+  justify-content: space-between;
+  padding: 18px;
+`
+
+const CloseIcon: FC<{
+  onClick?: () => void
+}> = ({ onClick }) => {
+  return (
+    <svg
```
delaaxe, line 18: I'd rather put all svgs in their own files

packages/extension/src/ui/components/Recovery/IconBar.tsx
```diff
@@ -0,0 +1,92 @@
...
+        />
+      ) : (
+        <div />
+      )}
+      <CloseIcon
+        onClick={() => {
+          navigate(routes.account())
+        }}
+      />
+    </IconBar>
```
delaaxe, line 90: spaghetti

packages/extension/src/ui/screens/ConfirmSeedRecovery.tsx
```diff
@@ -0,0 +1,233 @@
...
+      &:hover {
+        background-color: rgba(255, 255, 255, 0.4);
+      }
+    `}
+`
+
+function getRandomWord(excludeWordlist: string[]): string {
+  const excludeIndexes = excludeWordlist.map((word) =>
+    wordlists.en.getWordIndex(word),
+  )
```
delaaxe, line 78: be more functional when you can, have you tried:

```suggestion
  const excludeIndexes = excludeWordlist.map(wordlists.en.getWordIndex)
```

packages/extension/src/ui/screens/ConfirmSeedRecovery.tsx
```diff
@@ -0,0 +1,233 @@
...
+      ? selectedWords.filter((selectedWord) => selectedWord !== word)
+      : [...selectedWords, word]
+
+    setSelectedWords(updatedSelectedWords)
+
+    const expectedWords = wordlists.en
+      .split(seedPhrase)
+      .filter((word) => confirmSeedPhrase.includes(word))
+
+    console.log(expectedWords, updatedSelectedWords)
```
delaaxe, line 146: console.log

packages/extension/src/ui/screens/ConfirmSeedRecovery.tsx
```diff
@@ -0,0 +1,233 @@
...
+              }
+            }}
+            disabled={showError}
+          >
+            Confirm
+          </Button>
+          {showError && (
+            <FormError
+              style={{
+                marginTop: "8px",
```
delaaxe, line 222:
```suggestion
                marginTop: 8,
```

packages/extension/src/ui/screens/SetupSeedRecovery.tsx
```diff
@@ -0,0 +1,110 @@
...
+    () => getSeedPhrase(),
+    {
+      suspense: true,
+    },
+  )
+
+  return (
+    <SeedPhraseGrid>
+      {seedPhrase.split(" ").map((word, index) => (
+        <SeedWordBadge key={index}>{word}</SeedWordBadge>
```
delaaxe, line 64: it's generally a bad idea to use `index` as react key, why can't we use word as index here?

packages/extension/src/ui/screens/ConfirmSeedRecovery.tsx
```diff
@@ -0,0 +1,233 @@
...
+      suspense: true,
+    },
+  )
+
+  const { data: confirmSeedPhrase = [] } = useSWRImmutable(
+    "confirmSeedPhrase",
+    () => createConfirmationWordlist(wordlists.en.split(seedPhrase)),
+    {
+      suspense: true,
+    },
```
delaaxe, line 131:
```suggestion
    { suspense: true },
```

packages/extension/src/background/wallet.ts
janek26: this.isSessionOpen is the only relevant check here. If that's true, both of the following are true aswell.
It was just to trick the TS compiler. If there's a better way to do this let me know (`!` is forbidden by eslint)

packages/extension/src/ui/App.tsx
janek26: why? :D that would be a huge change to the entire codebase. I think both formats are readable

packages/extension/src/ui/components/Account/UpdateBanner.tsx
janek26: agree :)

packages/extension/src/ui/components/Recovery/IconBar.tsx
janek26: solved, thanks :)

packages/extension/src/ui/screens/ConfirmSeedRecovery.tsx
janek26: thanks

packages/extension/src/ui/App.tsx
janek26: We'll keep it for now in the settings for people to download :) We'll also keep the recovery by backup file for a few more versions

janek26: you can find it a little further down: 
```
<Route
  path={routes.backupDownload()}
  element={<BackupDownloadScreen />}
/>
```

packages/extension/src/ui/screens/SetupSeedRecovery.tsx
janek26: we need both as an id now, as the number (index) will display in the future

janek26:
waiting for #524

gergold (approved):

PR #542, opened 2022-04-25
feat(wallet): nfts

delaaxe:
NFT support


packages/extension/src/ui/utils/assertNever.ts
```diff
@@ -0,0 +1,3 @@
+export const assertNever = (_: never): never => {
```
janek26, line 1: this file does seems to be not used anymore. Can we remove it?

janek26 (approved):
great work! some minor things, otherwise approved!
But we should not release without syncing with Oasis, merging is probably fine already.

PR #546, opened 2022-04-29
feat(nfts): transfer nfts

delaaxe:
Send NFTs from one wallet to another


packages/extension/src/ui/screens/NftScreen.tsx
```diff
@@ -54,22 +64,44 @@ export const Container = styled.div`
...
+  if (!account || !nft || !contractAddress || !tokenId) {
     return <></>
   }
 
+  const handleSubmit = async (e: React.FormEvent) => {
+    e.preventDefault()
+
+    sendTransaction({
+      to: contractAddress,
+      method: "transferFrom",
```
janek26, line 94: as we don't call save transfer, should we at least check that the recipient is a valid address?

janek26 (approved):

PR #572, opened 2022-05-03
feat(ui): new disclaimer

delaaxe:


janek26:
Can we also please make the disable state of the button more obvious (change bg color)?
I can also scroll a tiny bit (1-5px). Can we make sure the content fits the fixed sized page, so the user can't scroll?

packages/extension/src/ui/screens/WelcomeScreen.tsx
```diff
@@ -8,6 +8,7 @@ import { Greetings, GreetingsWrapper } from "../components/Greetings"
 import { StickyArgentFooter } from "../components/StickyArgentFooter"
 import { P } from "../components/Typography"
 import { routes } from "../routes"
+import { isDisclaimerUnderstood } from "../utils/disclaimer"
```
janek26, line 11: can we remove this additional piece of logic? 
I think it is okay to show the disclaimer every time the user creates a new wallet

delaaxe: done

janek26 (approved):

PR #573, opened 2022-05-03
Feature/network-status

janek26:


packages/extension/src/background/swr/helpers.ts
```diff
@@ -0,0 +1,55 @@
...
+export const isFunction = (value: any): value is Function =>
+  typeof value === "function"
+
+type Nil = null | undefined
+
+export const isNil = (value: any): value is Nil =>
+  typeof value === "undefined" || value === null
+
+export const isPlainObject = (value: any) =>
+  !!value && typeof value === "object" && !Array.isArray(value)
```
delaaxe, line 13: we should rather use `lodash-es` than reimplement language helpers

packages/extension/src/background/swr/index.ts
```diff
@@ -0,0 +1,90 @@
...
+      }
+
+      return cachedValue as ReturnValue
+    }
+
+    return revalidate()
+  }
+
+  return staleWhileRevalidate
+}
```
delaaxe, line 90: why so much more caching code? caching is complicated and should rarely be implement in app

more code = more bugs

packages/extension/src/ui/components/NetworkSwitcher.tsx
```diff
@@ -118,12 +108,26 @@ export const NetworkSwitcher: FC<NetworkSwitcherProps> = ({ disabled }) => {
...
+
+  useEffect(() => {
+    if (
+      (networkStatuses[currentNetwork.id] === "degraded" ||
+        networkStatuses[currentNetwork.id] === "error") &&
+      needsToShowNetworkStatusWarning
+    ) {
+      navigate(routes.networkWarning())
+    }
+  }, [networkStatuses[currentNetwork.id]])
```
delaaxe, line 122: `networkStatuses[currentNetwork.id]` should be a temporary variable

packages/extension/src/ui/components/StatusIndicator.tsx
```diff
@@ -0,0 +1,96 @@
+import { FC } from "react"
+import styled, { css, keyframes } from "styled-components"
+
+import { NetworkStatus } from "../../shared/networks"
+import type { AccountStatusCode } from "../utils/accounts"
+import { NetworkWarning } from "./Icons/NetworkWarning"
+
+export type StatusIndicatorStatus = "green" | "orange" | "red" | "transparent"
```
delaaxe, line 8:
```suggestion
export type StatusIndicatorColor = "green" | "orange" | "red" | "transparent"
```

packages/extension/src/ui/components/StatusIndicator.tsx
```diff
@@ -0,0 +1,96 @@
...
+): StatusIndicatorStatus {
+  switch (status) {
+    case "CONNECTED":
+      return "green"
+    case "DEPLOYING":
+      return "orange"
+    case "ERROR":
+      return "red"
+    default:
+      return "transparent"
```
delaaxe, line 25: Could use `assertNever` here to have a compile error in case we add new statuses

packages/extension/src/ui/components/StatusIndicator.tsx
```diff
@@ -0,0 +1,96 @@
...
+  status?: NetworkStatus,
+): StatusIndicatorStatus {
+  switch (status) {
+    case "error":
+      return "red"
+    case "degraded":
+      return "orange"
+    case "ok": // network status shows green by default and has no transparent state
+    default:
+      return "green"
```
delaaxe, line 39: same comment (if it makes sense here)

packages/extension/src/ui/components/StatusIndicator.tsx
```diff
@@ -0,0 +1,96 @@
...
+import styled, { css, keyframes } from "styled-components"
+
+import { NetworkStatus } from "../../shared/networks"
+import type { AccountStatusCode } from "../utils/accounts"
+import { NetworkWarning } from "./Icons/NetworkWarning"
+
+export type StatusIndicatorStatus = "green" | "orange" | "red" | "transparent"
+
+interface StatusIndicatorProps {
+  status?: StatusIndicatorStatus
```
delaaxe, line 11:
```suggestion
  color?: StatusIndicatorStatus
```

packages/extension/src/ui/screens/NetworkWarningScreen.tsx
```diff
@@ -0,0 +1,56 @@
...
+        updateNeedsToShowNetworkStatusWarning()
+        navigate(routes.accounts())
+      }}
+    >
+      <Wrapper>
+        <Congestion />
+        <H2>Network issues</H2>
+        <CenterP>
+          StarkNet is in Alpha and is experiencing degraded network
+          performance.Your transactions may fail.
```
delaaxe, line 51:
```suggestion
          performance. Your transactions may fail.
```

packages/extension/src/ui/states/seenNetworkStatusWarning.ts
```diff
@@ -0,0 +1,31 @@
...
+        }),
+    }),
+    { name: "seenNetworkStatusState" },
+  ),
+)
+
+export const useNeedsToShowNetworkStatusWarning = () => {
+  const { lastSeen, updateLastSeen } = useSeenNetworkStatus()
+
+  console.log(lastSeen, Date.now())
```
delaaxe, line 25: console

delaaxe (changes requested):

packages/extension/src/background/networkStatus.ts
```diff
@@ -0,0 +1,109 @@
...
+  networks: Network[],
+): Promise<Partial<Record<Network["id"], NetworkStatus>>> => {
+  const statuses = await Promise.all(
+    networks.map((network) => getNetworkStatus(network)),
+  )
+
+  return networks.reduce(
+    (acc, network, i) => ({ ...acc, [network.id]: statuses[i] }),
+    {},
+  )
```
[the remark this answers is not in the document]
janek26: it would need to be something like:
```
return Object.fromEntries(statuses.map(ns => [ns.id, ns]))
```
Not sure it's more readable

packages/extension/src/ui/components/StatusIndicator.tsx
janek26: how exactly? I can just think of methods to catch it in runtime

packages/extension/src/background/swr/index.ts
janek26: I think swr is a very powerful pattern to
a) solve caching for network requests
b) solve loading states for the ui (I would not consider this network cache)

packages/extension/src/ui/components/StatusIndicator.tsx
delaaxe, line 25:
```suggestion
    case "DEFAULT":
      return "transparent"
    default:
      return assertNever(status)
```

packages/extension/src/ui/components/StatusIndicator.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+export const StatusIndicator = styled.span<StatusIndicatorProps>`
+  height: 8px;
+  width: 8px;
+  border-radius: 8px;
+
+  background-color: ${({ color = "transparent" }) =>
+    color === "green"
+      ? "#02BBA8"
+      : status === "orange"
```
delaaxe, line 59: should be color

janek26: good catch

delaaxe (approved):

PR #575, opened 2022-05-04
refactor(ui): begin with some feature folders

delaaxe:
Code co-location makes it easier to come back to a feature later on instead of having to scour multiple files sprinkled all over the repo


packages/extension/src/ui/App.tsx
```diff
@@ -5,6 +5,17 @@ import styled, { createGlobalStyle } from "styled-components"
 import { normalize } from "styled-normalize"
 import { SWRConfig } from "swr"
 
+import NftScreen from "./features/accountNfts/NftScreen"
```
janek26, line 8: can we add named export here? (as everywhere else)

janek26 (approved):

PR #582, opened 2022-05-05
feat: store tokens in background

janek26:
this allows to do checks on the token list without opening the UI
eg: checking if a token already exists in case a dapp wants to add one

### DIFF
--- a/packages/extension/src/background/tokens.ts
+++ b/packages/extension/src/background/tokens.ts
@@ -8,27 +8,21 @@ const tokenStore = new Storage<{
 }>({ tokens: [] }, "tokens")
 
 const validateToken = (token: Partial<Token>): token is Token => {
-  if (!token.address) {
-    return false
-  }
   try {
+    if (
+      !token.address ||
+      !token.networkId ||
+      !token.name ||
+      !token.symbol ||
+      !token.decimals
+    ) {
+      throw Error("token is missing required field")
+    }
     validateAndParseAddress(token.address)
+    return true
   } catch (e) {
     return false
   }
-  if (!token.networkId) {
-    return false
-  }
-  if (!token.name) {
-    return false
-  }
-  if (!token.symbol) {
-    return false
-  }
-  if (!token.decimals) {
-    return false
-  }
-  return true
 }
 
 export const getTokens = async (): Promise<Token[]> => {
@@ -40,11 +34,11 @@ export const hasToken = async (address: Token["address"]): Promise<boolean> => {
   return tokens.some((token) => token.address === address)
 }
 
-type MutateTokenRes =
+type TokenMutationResult =
   | { success: true; tokens: Token[] }
   | { success: false; tokens?: never }
 
-export const addToken = async (token: Token): Promise<MutateTokenRes> => {
+export const addToken = async (token: Token): Promise<TokenMutationResult> => {
   if (!validateToken(token)) {
     throw new Error("Invalid token")
   }
@@ -59,7 +53,7 @@ export const addToken = async (token: Token): Promise<MutateTokenRes> => {
 
 export const removeToken = async (
   tokenAddress: Token["address"],
-): Promise<MutateTokenRes> => {
+): Promise<TokenMutationResult> => {
   const tokens = await tokenStore.getItem("tokens")
   const index = tokens.findIndex(({ address }) => address === tokenAddress)
   if (index === -1) {
--- a/packages/extension/src/background/tokens.ts
+++ b/packages/extension/src/background/tokens.ts
@@ -20,7 +20,7 @@ const validateToken = (token: Partial<Token>): token is Token => {
     }
     validateAndParseAddress(token.address)
     return true
-  } catch (e) {
+  } catch {
     return false
   }
 }
--- a/packages/extension/src/background/tokens.ts
+++ b/packages/extension/src/background/tokens.ts
@@ -1,0 +1,71 @@
+import { validateAndParseAddress } from "starknet"
+
+import { Token } from "../shared/token"
+import { Storage } from "./storage"
+
+const tokenStore = new Storage<{
+  tokens: Token[]
+}>({ tokens: [] }, "tokens")
+
+const validateToken = (token: Partial<Token>): token is Token => {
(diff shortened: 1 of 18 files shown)


packages/extension/src/shared/token.ts
```diff
@@ -0,0 +1,36 @@
...
+  networkId: string
+  name: string
+  symbol: string
+  decimals: string
+  image?: string
+  showAlways?: boolean
+}
+
+export type RequestToken = Partial<
+  Omit<Token, "image" | "showAlways" | "address">
```
delaaxe, line 14: define a base interface which has the minimum number of fields and make `Token` extend that interface. no `Omit` here

packages/extension/src/shared/token.ts
```diff
@@ -0,0 +1,36 @@
...
+> &
+  Pick<Token, "address">
+
+export const equalToken = (
+  a: Pick<Token, "address" | "networkId">,
+  b: Pick<Token, "address" | "networkId">,
```
delaaxe, line 20: similarly, this is hacky

packages/extension/src/ui/states/tokens.ts
```diff
@@ -1,128 +1,124 @@
 import { BigNumber } from "ethers"
 import { useEffect, useMemo, useRef } from "react"
+import { number } from "starknet"
```
delaaxe, line 3: at some point we should change the pattern to just 
```typescript
import starknet from "starknet"
```
and use methods form there instead of importing very generic sounding names like `number`

packages/extension/src/ui/states/tokens.ts
```diff
@@ -1,128 +1,124 @@
...
 import { useSelectedAccount } from "./account"
 import { useAppState } from "./app"
 
-export interface TokenDetails {
-  address: string
-  name?: string
-  symbol?: string
+export interface TokenDetails
+  extends Pick<Token, "address" | "networkId">,
+    Partial<Omit<Token, "decimals" | "address" | "networkId">> {
```
delaaxe, line 22: see comment above

packages/extension/src/ui/states/tokens.ts
```diff
@@ -1,128 +1,124 @@
...
+    set((state) => ({
+      tokens: [...state.tokens, newToken],
+    }))
+  },
+  removeToken: async (tokenAddress: TokenDetails["address"]) => {
+    await removeTokenMsg(tokenAddress)
+
+    // optimistic update
+    set((state) => ({
+      tokens: state.tokens.filter((t) => t.address !== tokenAddress),
```
delaaxe, line 78: avoid single letter variables

packages/extension/src/ui/states/tokens.ts
```diff
@@ -1,128 +1,124 @@
...
+    }))
+  },
+}))
+
+export const useTokensSubscription = () => {
+  const { data: tokens = [] } = useSWRImmutable("tokens", () => getTokens(), {
```
delaaxe, line 84: if `getTokens` doesn't use the `this` keyword, this is better written as

```suggestion
  const { data: tokens = [] } = useSWRImmutable("tokens", getTokens, {
```

packages/extension/src/ui/states/tokens.ts
```diff
@@ -1,128 +1,124 @@
...
+}
 
-export const addToken = (token: TokenDetails) => {
+export const addToken = (token: Required<TokenDetails>) => {
   useTokens.getState().addToken(token)
 }
 
-export const removeToken = (token: TokenDetails) => {
-  useTokens.getState().removeToken(token)
+export const removeToken = (tokenAddress: TokenDetails["address"]) => {
```
delaaxe, line 115: can't just use `tokenAddress: string`?

packages/extension/src/ui/utils/messaging.ts
```diff
@@ -233,3 +234,25 @@ export const getNetworkStatuses = async (networks: Network[] = []) => {
...
   return waitForMessage("GET_NETWORK_STATUSES_RES")
 }
+
+//   | { type: "GET_TOKENS" }
+//   | { type: "GET_TOKENS_RES"; data: Token[] }
+//   | { type: "UPDATE_TOKENS"; data: Token[] }
+//   | { type: "REMOVE_TOKEN"; data: Token["address"] }
+//   | { type: "REMOVE_TOKEN_RES"; data: boolean }
+//   | { type: "ADD_TOKEN"; data: Required<Token> }
+//   | { type: "ADD_TOKEN_RES"; data: boolean }
```
delaaxe, line 244: commented code

delaaxe (changes requested):

packages/extension/src/background/tokens.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+  if (!token.name) {
+    return false
+  }
+  if (!token.symbol) {
+    return false
+  }
+  if (!token.decimals) {
+    return false
+  }
```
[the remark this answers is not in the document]
janek26: yes sorry, evolved into this monster :D

packages/extension/src/ui/states/tokens.ts
janek26: the name of that export is being worked on in the starknet.js group currently

packages/extension/src/ui/states/tokens.ts
janek26: it's equivallent. With the benefit that this type will change when the interface changes :) helps to catch errors while refactors imo

delaaxe: I doubt the address type will change, and even if it does there will be plenty of compile errors everywhere. `string` is more readable and easier to understand

janek26: done :)

packages/extension/src/shared/token.ts
janek26: It ensures you need to pass as little data as possible, I think that's a good pattern. 
It allows you to compare Tokens with BigInt decimals with Tokens with number decimals ie

delaaxe (approved):

PR #585, opened 2022-05-05
refactor(ui): prefer using `Link`s to `useNavigate`

delaaxe:


packages/extension/src/ui/App.tsx
```diff
@@ -69,6 +69,11 @@ const GlobalStyle = createGlobalStyle`
     padding: 0;
     margin-block: 0;
   }
+
+  a {
+    text-decoration: none;
+    color: inherit;
```
janek26, line 75: I remember if you want to reset links you also need to handle the visited state, isn't that true anymore?
https://developer.mozilla.org/en-US/docs/Web/CSS/:visited

packages/extension/src/ui/components/Account/AccountContainer.tsx
```diff
@@ -57,24 +55,25 @@ export const AccountContainer: FC<AccountScreenContentProps> = ({
     <Container header footer>
       <AccountHeader>
         <Header>
-          <ProfilePicture
-            {...makeClickable(() => navigate(routes.accounts()))}
-            src={getAccountImageUrl(accountName, account.address)}
-          />
+          <Link to={routes.accounts()}>
```
janek26, line 58: Link does not replace makeClickable, makeClickable adds a11y and focus navigation. Does it work the same with Link?

packages/extension/src/ui/App.tsx
delaaxe: seems good enough: https://stackoverflow.com/questions/8919682/remove-all-styling-formatting-from-hyperlinks

packages/extension/src/ui/components/Account/AccountContainer.tsx
delaaxe: idk, do people really use alt-tabbing and stuff like that on crypto extensions? I don't see a difference

packages/extension/src/ui/App.tsx
janek26: thanks for checking

packages/extension/src/ui/components/Account/AccountContainer.tsx
janek26: it makes the extension useable with a screen reader

delaaxe: that makeClickable pattern is quite cumbersome and I haven't seen it used elsewhere. if we really need to use it can we wrap it into a ClickableLink or something in a another release?

janek26: yeah, some custom link component to keep the a11y navigation would be super!

janek26 (approved):
I just checked this PR out and it seems like a tags add a11y themselves :) The active state looks a little different/weird, but fine for now

PR #625, opened 2022-05-12
feat: move transactions to background

janek26:
by moving all transactions logic to the background process, we're enabling multiple sources for transaction details and status.
In this implementation we track all transactions which are made through the extension and update the state from onchain as long as it's not final (reached ACCEPTED_ON_L2)
In case all cases including and after ACCEPTED_ON_L2 it should also show in the voyager history api, and once that's the case we update the details.
This allows for much better transaction tracking, history overview, transaction naming and much more.


packages/extension/src/shared/transactions.ts
```diff
@@ -0,0 +1,20 @@
+import { Status } from "starknet"
+
+import { WalletAccount } from "./wallet.model"
+
+export interface TransactionMeta {
+  title?: string
+  subTitle?: string
+}
+
+export interface AddTransaction {
```
delaaxe, line 10: class names and file names shouldn't start with a verb

```suggestion
export interface TransactionRequest {
```

packages/extension/src/ui/features/accounts/accountTransactions.state.ts
```diff
@@ -2,15 +2,35 @@ import { useEffect } from "react"
...
 import { getTransactions } from "../../services/messaging"
 
 interface State {
-  transactions: TransactionStatus[]
+  transactions: Transaction[]
+  addTransactions: (transactions: Transaction[]) => void
 }
 
-const useTransactionsStore = create<State>(() => ({
+function mergeTransactionsArrays(
```
delaaxe, line 13:
```suggestion
function mergeTransactionArrays(
```

packages/extension/src/ui/features/accounts/accountTransactions.state.ts
```diff
@@ -2,15 +2,35 @@ import { useEffect } from "react"
...
+  transactions: Transaction[],
+  newTransactions: Transaction[],
+): Transaction[] {
+  return [
+    ...transactions.filter(
+      (transaction) =>
+        !newTransactions.find((t) => t.hash === transaction.hash),
+    ),
+    ...newTransactions,
+  ]
```
delaaxe, line 23: not like this?
```suggestion
  return uniqBy([...transactions, ...newTransactions], ({ hash }) => hash)
```

packages/extension/test/utils.ts
```diff
@@ -0,0 +1,17 @@
+import { IStorage } from "../src/background/storage"
```
delaaxe, line 1: file should be named `mocks.ts`, no utils/helpers/tools/things/stuff etc.

packages/extension/webpack.config.js
```diff
@@ -77,12 +77,14 @@ module.exports = {
...
-    },
-  },
+  optimization: isProd
+    ? {
+        minimize: true,
+        splitChunks: {
+          chunks: "async",
+        },
+      }
+    : undefined,
```
delaaxe, line 87:
```suggestion
  optimization: isProd && {
    minimize: true,
    splitChunks: { chunks: "async" },
  },
```

packages/extension/src/background/index.ts
```diff
@@ -633,8 +632,10 @@ import { Wallet, WalletStorageProps } from "./wallet"
         const network = msg.data
         try {
           const { account, txHash } = await wallet.addAccount(network)
-          transactionTracker.trackTransaction(txHash, account, {
-            title: "Deploy wallet",
+          transactionTracker.addAccount(account, {
+            hash: txHash,
+            account,
```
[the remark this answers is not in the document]
janek26: yes, you could imagine 2 different accounts in both places in the future.
The transaction also takes metadata.

packages/extension/src/background/storage/array.ts
```diff
@@ -0,0 +1,52 @@
...
+    const inner = await this.store.getItem("inner")
+    return inner.filter(filterFn)
+  }
+
+  async getItem(findFn: (item: T) => boolean): Promise<T | null> {
+    const inner = await this.store.getItem("inner")
+    return inner.find(findFn) ?? null
+  }
+
+  async addItems(values: T[]): Promise<void> {
```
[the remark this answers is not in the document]
janek26: found a good solution with `uniqWith`, thanks

packages/extension/src/background/transactions/onchain.ts
```diff
@@ -0,0 +1,26 @@
...
+      const status = await provider.getTransactionStatus(transaction.hash)
+      return {
+        ...transaction,
+        status: status.tx_status,
+        failureReason: status.tx_failure_reason,
+      }
+    }),
+  )
+
+  const updatedTransactions = fetchedTransactions.reduce((acc, transaction) => {
```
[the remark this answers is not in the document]
janek26: whilst this was true for the other case, this one is different because of TS typing of SettledPromise

packages/extension/src/background/transactions/transactions.ts
```diff
@@ -0,0 +1,102 @@
...
+    const historyUpdates = getTransactionsStatusUpdate(
+      allTransactions,
+      historyTransactions,
+    )
+
+    const updates = [...onchainUpdates, ...historyUpdates]
+
+    onUpdate?.(updates)
+
+    return transactionsStore.addItems(updates)
```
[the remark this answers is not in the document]
janek26: yep, added, thanks :)

packages/extension/src/shared/transactions.ts
```diff
@@ -0,0 +1,20 @@
...
+  hash: string
```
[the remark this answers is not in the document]
janek26: would likely result in a new type in DailyActivity and would not be related to Transactions

packages/extension/src/ui/features/accounts/accountTransactions.state.ts
janek26: thanks, dont know lodash well

packages/extension/test/transactions.test.ts
```diff
@@ -0,0 +1,224 @@
+import "isomorphic-fetch"
```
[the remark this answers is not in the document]
janek26: agree, but the test folder is the current pattern, so I kept it

packages/extension/webpack.config.js
janek26: false doesnt match the webpack config format, therefore it needs to be undefined

packages/extension/src/background/index.ts
```diff
@@ -95,22 +97,18 @@ import { Wallet, WalletStorageProps } from "./wallet"
...
+            data: tracked,
           })
         }
 
-        const provider = await getProvider(msg.data.network)
-        const fetchedStatus = await getTransactionStatus(
-          provider,
-          msg.data.hash,
-        )
+        // lets keep this for easier debugging please
```
delaaxe, line 108: then it should be bubbled back into the UI or pushed to an analytics service

packages/extension/src/background/transactions/transactions.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    const updates = [...onChainUpdates, ...historyUpdates]
+
+    try {
+      onUpdate?.(updates)
+    } catch {
+      // ignore
+    }
+
+    return transactionsStore.addItems(updates)
```
delaaxe, line 74:
```suggestion
    transactionsStore.addItems(updates)
    onUpdate?.(updates)
```

packages/extension/src/background/wallet.ts
```diff
@@ -87,20 +87,27 @@ function calculateContractAddress(
   ])
 }
 
+export const equalAccount = (
+  a: Pick<WalletAccount, "address" | "network">,
+  b: Pick<WalletAccount, "address" | "network">,
+) => a.address === b.address && a.network.id === b.network.id
+
+export type GetNetworkFunction = (networkId: string) => Promise<Network>
```
delaaxe, line 95:
```suggestion
export type GetNetwork = (networkId: string) => Promise<Network>
```

packages/extension/src/shared/transactions.ts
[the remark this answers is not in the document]
delaaxe: cool

packages/extension/webpack.config.js
delaaxe: it works for me this way

delaaxe (changes requested):

packages/extension/webpack.config.js
janek26: Bildschirmfoto 2022-05-12 um 23 15 39

delaaxe (approved):

PR #638, opened 2022-05-13
feat(settings): request password before showing seed

delaaxe:


packages/extension/src/ui/features/onboarding/PasswordForm.tsx
```diff
@@ -0,0 +1,64 @@
...
+interface PasswordFormProps {
+  verifyPassword: (password: string) => Promise<boolean>
+  children?: (isDirty: boolean) => ReactNode
+}
+
+export const PasswordForm: FC<PasswordFormProps> = ({
+  verifyPassword,
+  children,
+}) => {
+  const [passwordError, setPasswordError] = useState<string>()
```
janek26, line 21: duplicated state

packages/extension/src/ui/features/onboarding/PasswordForm.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const handlePassword: SubmitHandler<FieldValues> = async ({ password }) => {
+    setPasswordError(undefined)
+    if (!(await verifyPassword(password))) {
+      setPasswordError("Wrong password")
+    }
+  }
+
+  return (
+    <form onSubmit={handleSubmit(handlePassword)}>
+      <Controller
```
janek26, line 38: maybe this should be a stateless component `PasswordInput` which accepts an optional error object render
that way we could remove the render props and allow for a more general use inside of any form component

janek26 (approved):
good work :) 

I wonder if we should add a timeout, of lets say 10 mins, after each session start, to allow the backup flow without password and show the password prompt once the timeout expired.

That way we could also protect the "Backup Seed Phrase"-Banner flow

janek26 (changes requested):
lets wait for a better request password screen design

packages/extension/src/ui/features/onboarding/PasswordForm.tsx
delaaxe: How, what's the duplicate?

packages/extension/src/ui/features/onboarding/PasswordForm.tsx
delaaxe: But then we'd need to duplicate th

delaaxe: But then we'd need to duplicate that whole `react-hook-forms` rigmarole which doesn't work well with the ConfirmScreen, unless you a better way

packages/extension/src/ui/features/onboarding/PasswordForm.tsx
delaaxe: How?

janek26 (approved):

PR #670, opened 2022-05-24
fix: populate tx history on recover

janek26:


delaaxe:
seems like a genuine test failure in CI

delaaxe (approved):

PR #672, opened 2022-05-24
fix: use better maxFee

janek26:


packages/extension/src/ui/features/actions/FeeEstimation.tsx
```diff
@@ -73,21 +73,26 @@ type FeeEstimationProps = (
...
+      if ("maxFee" in x) {
+        return { unit: "wei", amount: x.maxFee, suggestedMaxFee: x.maxFee }
+      }
+      return getEstimatedFee(x)
+    },
+    {
+      suspense: true,
+      refreshInterval: 20 * 1000, // 20 seconds
+    },
+  )
```
delaaxe, line 95: elaborate logic like this should be pulled outside of components

packages/extension/src/shared/MessageType.ts
```diff
@@ -30,6 +30,7 @@ export type MessageType =
       data: {
         amount: string
         unit: string
+        suggestedMaxFee: string
```
delaaxe, line 33: ESTIMATE_TRANSACTION_* should be under the ***** transactions ***** section

delaaxe (approved):

PR #690, opened 2022-05-30
feat(background): use constructor-initialized proxy

delaaxe:


delaaxe:
to be merged with cairo 0.9 support

janek26:
This should differentiate between Mainnet and Testnet, as we still need the old version on main net until 0.9.0 releases on mainnet

janek26 (changes requested):
This should differentiate between Mainnet and Testnet, as we still need the old version on main net until 0.9.0 releases on mainnet

packages/extension/src/background/wallet.ts
```diff
@@ -360,6 +363,77 @@ export class Wallet {
 
   public async addAccount(
     networkId: string,
+  ): Promise<{ account: WalletAccount; txHash: string }> {
+    // FIXME: delete this once Cairo 9 is on mainnet
+    if (networkId === "mainnet-alpha") {
```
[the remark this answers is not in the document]
delaaxe: TS2367: This condition will always return 'true' since the types '"goerli-alpha"' and '"integration"' have no overlap.

janek26:
please keep using the old compiled contract versions for the old deployment flow

janek26:
please keep using the old compiled contract versions for the old deployment flow

packages/extension/src/ui/features/settings/NetworkSettingsFormScreen.tsx
```diff
@@ -46,14 +47,11 @@ type NetworkSettingsFormScreenProps =
...
-      return {
-        id: "",
-        name: "",
-        chainId: "",
-        baseUrl: "",
-      }
+      return { id: "", name: "", chainId: "", baseUrl: "" }
     }
     return props.network
   }, [props])
```
delaaxe, line 57: @janek26 `props` is a different reference on every render so `useMemo` effectively does nothing here and potentially consumes a bunch of extra memory, am I wrong?

janek26: right, should be 
```suggestion
  }, [props.network, props.mode])
```

janek26 (approved):

PR #693, opened 2022-05-31
fix(ui): add breathing space for deploying status

delaaxe:


janek26 (approved):
big change - next time please split it up into multiple commits so it's easier to review!

PR #699, opened 2022-06-02
feat: Hide accounts

dhruvkelawala:


packages/extension/src/ui/features/accountTokens/AccountMenu.tsx
```diff
@@ -104,6 +113,17 @@ export const AccountMenu: FC<AccountNameProps> = ({ onAccountNameEdit }) => {
             </MenuItem>
           </MenuItemWrapper>
           <Separator />
+          {account && (
+            <MenuItemWrapper onClick={() => hideAccount(account)}>
+              <MenuItem>
+                <IconWrapper>
+                  <VisibilityOff fontSize={"inherit"} htmlColor={"white"} />
```
delaaxe, line 120:
```suggestion
                  <VisibilityOff fontSize="inherit" htmlColor="white" />
```

packages/extension/src/ui/features/accountTokens/useHideAccountCallback.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { routes } from "../../routes"
+import { deleteAccount } from "../../services/messaging"
+import { Account } from "../accounts/Account"
+import { useAccounts } from "../accounts/accounts.state"
+
+export const useHideAccountCallback = () => {
+  const { hideAccount } = useAccounts()
+  const navigate = useNavigate()
+
+  return useCallback(async (account: Account) => {
```
delaaxe, line 13: is useCallback really needed? honest question

packages/extension/src/ui/features/accounts/accounts.state.ts
```diff
@@ -3,9 +3,10 @@ import create from "zustand"
 import { Account } from "./Account"
 
 interface State {
-  accounts: Record<string, Account>
+  accounts: { [address: string]: Account }
```
delaaxe, line 6:
```suggestion
  accounts: Record<string, Account>
```

delaaxe (commented):
is there a figma for this? why is the UX different than for hiding localhost accounts?

packages/extension/src/ui/features/accounts/accounts.state.ts
dhruvkelawala: Cannot use spread operator on `Record<string, Account>` and we are using spread operator in `hideAccount`

delaaxe (commented):
great. one last question: after restarting the browser, the hidden accounts reappear. is this expected?

dhruvkelawala:
No it isn't expected and it isn't happening on my end. But we are updating logic anyways where we won't remove the Account from the storage but just add it to the list of HiddenAccounts as discussed with @juniset

janek26:
as far as I understood this is not ready for review yet. Correct me if I'm wrong

packages/extension/src/background/accountMessaging.ts
```diff
@@ -17,7 +17,7 @@ export const handleAccountMessage: HandleMessage<AccountMessage> = async ({
     case "GET_ACCOUNTS": {
       return sendToTabAndUi({
         type: "GET_ACCOUNTS_RES",
-        data: await wallet.getAccounts(),
+        data: (await wallet.getAccounts()).filter((acc) => !acc.hidden),
```
janek26, line 20: I think getAccounts default behavior should be to only show non hidden accounts.
+ there should be an optional argument to show hidden accounts if you want to:
```
function getAccounts(showHidden = false): Accounts[]
```

packages/extension/src/background/accountMessaging.ts
```diff
@@ -116,6 +116,18 @@ export const handleAccountMessage: HandleMessage<AccountMessage> = async ({
       }
     }
 
+    case "HIDE_ACCOUNT": {
+      try {
+        await wallet.hideAccount(msg.data)
+
+        console.log("Account Hide Successful")
```
janek26, line 123: remove console log before going to prod

janek26 (commented):
after latest merges there are also conflicts

janek26:
please rebase/merge develop

packages/extension/src/background/wallet.ts
```diff
@@ -417,6 +433,7 @@ export class Wallet {
         type: "local_secret",
         derivationPath: getPathForIndex(index),
       },
+      hidden: false,
```
delaaxe, line 436: idem

packages/extension/src/background/accountMessaging.ts
dhruvkelawala: I also had the similar idea, but on most of the places where we are using getAccounts(), we want all the accounts including the hidden ones. That's why I thought it was a better idea to filter them here.

packages/extension/src/ui/features/accounts/accounts.state.ts
```diff
@@ -18,6 +21,11 @@ export const useAccounts = create<State>((set) => ({
         [newAccount.address]: newAccount,
       },
     })),
+  hiddenAccounts: [],
+  hideAccount: (account: Account) =>
+    set((state) => ({
+      hiddenAccounts: [...state.hiddenAccounts, account.address],
+    })),
```
delaaxe, line 28: why are changes needed in UI state if everything is handled in background? @janek26

packages/extension/src/ui/services/messaging.ts
```diff
@@ -193,6 +193,21 @@ export const deleteAccount = async (address: string) => {
...
+  sendMessage({ type: "HIDE_ACCOUNT", data: address })
+
+  try {
+    await Promise.race([
+      waitForMessage("HIDE_ACCOUNT_RES"),
+      waitForMessage("HIDE_ACCOUNT_REJ").then(() => {
+        throw new Error("Rejected")
+      }),
+    ])
+  } catch (error) {
```
delaaxe, line 206:
```suggestion
  } catch {
```

packages/extension/src/background/wallet.ts
```diff
@@ -280,6 +295,7 @@ export class Wallet {
                 type: "local_signer",
                 derivationPath: getPathForIndex(lastCheck),
               },
+              hidden: false,
```
delaaxe, line 298: should not be if `hidden` is `boolean | undefined`

```suggestion
```

delaaxe (changes requested):

packages/extension/src/ui/features/accounts/accounts.state.ts
dhruvkelawala: Because AccountList is fetched from the UI state, not the background. Accounts are fetched from the background only once after the extension is reloaded or browser is restarted

packages/extension/src/background/accountMessaging.ts
delaaxe: I agree with Janek, his suggestion seems cleaner

delaaxe (approved):

janek26:
Hey, after rebasing this PR I think we should change several things:
1. Don't have multiple source of truth, the background should hold the only state that determines if an account is hidden or not. Please remove `hiddenAccounts` from https://github.com/argentlabs/argent-x/blob/feat/hide-accounts/packages/extension/src/ui/features/accounts/accounts.state.ts#L24
2. getAccounts should default to not returning hidden accounts. That way the UI will fully ignore them.
3. useAccounts accounts object can be kept in sync with background messages like you can find here: https://github.com/argentlabs/argent-x/blob/86849c6a07d1ed7ff46121a97539b6e139db81a7/packages/extension/src/ui/features/accountTokens/tokens.state.ts

I know it's done differently with `addAccount` on the same state object, but long term that should move into a similar pattern

delaaxe:
1. I had the exact same thought on multiple sources of truth but forgot to mention it
2. Keep in mind knowing hidden accounts may be needed to determine if MigrationDisclaimerScreen needs to be show

delaaxe (changes requested):
1. I had the exact same thought on multiple sources of truth but forgot to mention it
2. Keep in mind knowing hidden accounts may be needed to determine if MigrationDisclaimerScreen needs to be show

dhruvkelawala:
Not ready for merge. Needs testing. 

Bug 1: Adding a new account and reloading extension, unhides and shows all the hidden account. Not everytime, sometimes only.

Bug 2: Sometimes adding a new account, recovers the old hidden account instead of creating a new account.

packages/extension/src/background/accountMessaging.ts
```diff
@@ -84,6 +84,19 @@ export const handleAccountMessage: HandleMessage<AccountMessage> = async ({
...
 
+    case "HIDE_ACCOUNT": {
+      try {
+        await wallet.hideAccount(msg.data)
+        const updatedAccounts = await wallet.getAccounts()
+        return sendToTabAndUi({
+          type: "HIDE_ACCOUNT_RES",
+          data: updatedAccounts,
+        })
+      } catch (error) {
```
delaaxe, line 95: Unused variable

packages/extension/src/background/wallet.ts
```diff
@@ -154,13 +154,18 @@ export class Wallet {
     this.setSession(ethersWallet.privateKey, password)
   }
 
-  public async getAccounts(): Promise<WalletAccount[]> {
+  public async getAccounts(showHidden = false): Promise<WalletAccount[]> {
```
delaaxe, line 157: Shouldn't the default be true?

This is not a display method, so "show" isn't the right term. "includeHidden" more appropriate

packages/extension/src/ui/features/accountTokens/AccountMenu.tsx
```diff
@@ -75,13 +84,20 @@ export const AccountMenu: FC<AccountNameProps> = ({ onAccountNameEdit }) => {
...
   useOnClickOutside(ref, () => setMenuOpen(false))
 
   const handleEditClick = () => {
     setMenuOpen(false)
     onAccountNameEdit()
   }
 
+  const handleHideAccount = async (account: Account) => {
+    await hideAccount(account.address)
+    navigate(routes.accounts())
```
delaaxe, line 98: Can you try navigate(await recover({ showList: true })? Should fix some of the bugs

packages/extension/src/ui/features/accounts/accounts.state.ts
```diff
@@ -33,3 +38,47 @@ export const useSelectedAccount = () =>
...
+      }
+
+      return () => {
+        if (!subscription.closed) {
+          subscription.unsubscribe()
+        }
+      }
+    })
+  }, [])
+}
```
delaaxe, line 84: All this may be unnecessary if you use the recover method from above

packages/extension/src/ui/services/backgroundAccounts.ts
```diff
@@ -39,6 +39,21 @@ export const deleteAccount = async (address: string) => {
...
+    await Promise.race([
+      waitForMessage("HIDE_ACCOUNT_RES"),
+      waitForMessage("HIDE_ACCOUNT_REJ").then(() => {
+        throw new Error("Rejected")
+      }),
+    ])
+  } catch {
+    throw Error("Could not hide account")
+  }
+}
```
delaaxe, line 55: Remove the try/catch, it's redundant because already throwing in REJ case

delaaxe (changes requested):

packages/extension/src/background/wallet.ts
dhruvkelawala: See the suggestion by Janek 

https://github.com/argentlabs/argent-x/pull/699#discussion_r888777127

packages/extension/src/ui/features/accounts/accounts.state.ts
dhruvkelawala: I used @janek26's suggestion of keeping the state account in sync with background using messaging

delaaxe: Is this still needed after the recover change?

packages/extension/src/background/wallet.ts
delaaxe: Naming unresolved

packages/extension/src/ui/features/accounts/accounts.state.ts
dhruvkelawala: Removed

packages/extension/src/ui/features/accountTokens/AccountMenu.tsx
dhruvkelawala: Done

packages/extension/src/background/wallet.ts
dhruvkelawala: Done

delaaxe:
we should deactivate the delete account now that we can hide: https://github.com/argentlabs/argent-x/blob/develop/packages/extension/src/ui/features/accounts/AccountListScreen.tsx#L119

delaaxe:
julien recommended removing it altogether for deprecated accounts and moving it to account menu for localhost accounts (because could be easily clicked by mistake as is) but up to you depending on the load

janek26:
makes sense, but probably not as part of this PR

janek26:
we agreed to
1. move "delete account" from account list into the account menu
2. always just show one of "delete account" or "hide account", depending on if an account is deprecated or on localhost.
3. always show a confirm screen in front of "delete account" and "hide account" action

janek26 (approved):

delaaxe (approved):

PR #726, opened 2022-06-03
refactor(ui): messaging split

delaaxe:


janek26:
think this is good, but I would remove the `background` prefix in the files name as it's already in the background folder

delaaxe:
no this pr is meant to be only about ui changes, is there a bug in `background/customNetworks`? it should not import ui files

janek26:
oh sorry misunderstood it. I think it probably should be a folder, like `ui/services/background/index.ts` etc

janek26 (approved):

PR #732, opened 2022-06-04
feat: new recovery

janek26:
if accountClassHash is provided for a network we use the new recovery mechanism. This allows only to recover new accounts and ignores any old accounts on that network.

If the network still has no accountClassHash we consider it pre cairo v0.9.0 and use the old recovery mechanism.

Pre cairo v0.9.0 should be removed as soon as devnet and mainnet support it.

### DIFF
--- a/packages/extension/src/background/recoveryMessaging.ts
+++ b/packages/extension/src/background/recoveryMessaging.ts
@@ -55,7 +55,8 @@ export const handleRecoveryMessage: HandleMessage<RecoveryMessage> = async ({
         transactionTracker.load(await wallet.getAccounts())
 
         return sendToTabAndUi({ type: "RECOVER_SEEDPHRASE_RES" })
-      } catch {
+      } catch (e) {
+        console.error(e)
         return sendToTabAndUi({ type: "RECOVER_SEEDPHRASE_REJ" })
       }
     }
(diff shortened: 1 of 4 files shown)


packages/extension/src/background/recoveryMessaging.ts
```diff
@@ -55,7 +55,8 @@ export const handleRecoveryMessage: HandleMessage<RecoveryMessage> = async ({
         transactionTracker.load(await wallet.getAccounts())
 
         return sendToTabAndUi({ type: "RECOVER_SEEDPHRASE_RES" })
-      } catch {
+      } catch (e) {
```
delaaxe, line 58: I would always use `error` to avoid single letter vars

packages/extension/src/background/wallet.ts
```diff
@@ -232,34 +246,125 @@ export class Wallet {
...
+
+        while (lastHit + offset > lastCheck) {
+          const starkPair = getStarkPair(
+            lastCheck,
+            secret,
+            newBaseDerivationPath,
+          )
+          const starkPub = ec.getStarkKey(starkPair)
+
+          const address = calculateContractAddress(
```
delaaxe, line 306: Eventually we could move all contract / class hash related to a contracts.ts to unclog this wallet.ts

delaaxe (approved):
Looks good

packages/extension/src/background/wallet.ts
janek26: I think it's just this one method, isn't it? 
I should transfer the method to starknet.js tbh

PR #737, opened 2022-06-07
fix: ensure normalized address in view on voyager link

arkoc:
image

- If the address has padded 0's, View On Voyager link is opening without initial 0's, which is fine for Voyager but it's creating inconsistency of address lengths and it can be confusing for customers. 
- For Copy to Clipboard it was already implemented (the value copied to the clipboard is padded with 0's)

### DIFF
--- a/packages/extension/src/ui/services/voyager.service.ts
+++ b/packages/extension/src/ui/services/voyager.service.ts
@@ -1,6 +1,7 @@
 import join from "url-join"
 
 import { Network } from "../../shared/networks"
+import { normalizeAddress } from "./addresses"
 
 export const getVoyagerContractLink = (
   address: string,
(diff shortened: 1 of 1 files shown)


janek26:
afaiu, we need the additional zero as soon as checksum addresses will become a thing. So I think we should keep them.

arkoc:
> afaiu, we need the additional zero as soon as checksum addresses will become a thing. So I think we should keep them.

copy address is adding 0's - copy voyager is not adding 0's
It's just simple inconsistency which is leading to confusing among users

janek26:
agree. But as I explained it will be used everywhere soon when checksum addresses are used. 

Voyager should add checksum addressses/64bit long addresses

janek26:
I try to reach out to voyager to show addresses with `0x0` and in a checksum format

PR #752, opened 2022-06-10
fix: transaction queuing

janek26:
and more, see individual commits


packages/extension/src/background/transactions/transactions.ts
```diff
@@ -77,6 +78,21 @@ export const getTransactionsTracker: GetTransactionsTracker = (
...
+      return // dont throw
+    }
+
+    const newTransaction = {
+      status: "RECEIVED" as const,
+      timestamp: timestampInSeconds(),
+      ...transaction,
+    }
+    onUpdate?.([newTransaction])
+    return transactionsStore.addItem(newTransaction)
```
delaaxe, line 93: As last time, isn't it better to do onUpdate last, in case it throws?

delaaxe (approved):

PR #755, opened 2022-06-12
feat: improve send input validation

janek26:


packages/extension/src/ui/features/accountTokens/TokenScreen.tsx
```diff
@@ -98,17 +99,24 @@ export const TokenScreen: FC = () => {
...
+        calldata: {
+          recipient,
+          amount: getUint256CalldataFromBN(
+            ethers.utils.parseUnits(amount, decimals),
+          ),
+        },
+      })
+      navigate(routes.accountTokens())
+    } catch {
+      // do nothing
```
delaaxe, line 118: can we show relevant error instead of doing nothing?

packages/extension/src/shared/token.ts
```diff
@@ -33,3 +37,42 @@ export const getFeeToken = (networkId: string) =>
     ({ symbol, networkId: network }) =>
       symbol === "ETH" && network === networkId,
   )
+
+export const parseAmount = (
+  amount: string,
+  decimals: number.BigNumberish = 18,
+) => {
+  return utils.parseUnits(
+    amount.replace(",", "."),
```
delaaxe, line 46: commas can sometimes be thousands separators but maybe we don't need to think about this

delaaxe (approved):

packages/extension/src/shared/token.ts
janek26: is that only for displaying purposes? or are people typing that way?

delaaxe: could be copy pasted from somewhere but nvm

janek26: maybe good to keep in mind when we redesign the page

PR #759, opened 2022-06-13
feat: shortcut to extended view

simonheys:
- Adds an 'Extended view' option to the Settings screen which opens the extension in a browser tab
- When in a tab, there are responsive horizontal margins (configurable, looks v weird without these)
- Handles the case where an external action is received to focus the extension if running in a tab

Screenshot 2022-06-13 at 16 44 55

Screenshot 2022-06-13 at 16 55 00

Screenshot 2022-06-13 at 16 55 08


packages/extension/src/ui/features/browser/tabs.ts
```diff
@@ -0,0 +1,38 @@
...
+import browser from "webextension-polyfill"
+
+export const openExtensionInTab = async () => {
+  const url = browser.runtime.getURL("index.html")
+  const tab = await browser.tabs.create({ url })
+  return tab
+}
+
+export const extensionIsInTab = async () => {
+  return Boolean(await browser.tabs.getCurrent())
```
janek26, line 11: sure getCurrent works as you expect? I would expect that it returns whatever tab is currently open by the window, eg google

packages/extension/src/ui/features/settings/SettingsScreen.tsx
```diff
@@ -70,75 +74,91 @@ const Footer = styled.div`
...
+            <ArrowForwardIosIcon fontSize="inherit" />
+          </Title>
+        </SettingsItem>
+        <hr />
+        {!extensionIsInTab && (
+          <>
+            <SettingsItem to={routes.settings()} onClick={openExtensionInTab}>
+              <Title>
+                <span>Extended view</span>
+                <OpenInFullIcon fontSize="inherit" />
```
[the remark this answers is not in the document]
simonheys: Changed to same chevron icon as the others

packages/extension/src/ui/features/actions/ActionScreen.tsx
```diff
@@ -1,12 +1,14 @@
-import { FC, useCallback } from "react"
+import { FC, useCallback, useEffect } from "react"
 import { useNavigate } from "react-router-dom"
+import browser from "webextension-polyfill"
```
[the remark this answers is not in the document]
simonheys: Good catch - changed

packages/extension/src/ui/features/browser/tabs.ts
simonheys: https://developer.chrome.com/docs/extensions/reference/tabs/#method-getCurrent

> Gets the tab that this script call is being made from

It does seem to work as expected. I think ideally there would be some UI tests around this feature, not sure if that is possible with web extensions?

janek26: ohh, cool to know!

e2e test are to be explored

janek26 (approved):

PR #764, opened 2022-06-15
feat: transaction review fallback screen

simonheys:
Adds client-side transaction review fallback screen with additional specific UI for ERC20 transfer

Screenshot 2022-06-15 at 10 04 50

Screenshot 2022-06-15 at 10 05 16

Screenshot 2022-06-15 at 10 05 19

### DIFF
--- a/packages/extension/src/shared/call.ts
+++ b/packages/extension/src/shared/call.ts
@@ -24,7 +24,7 @@ export interface Erc20TransferCall extends Call {
  * @returns true if the call passes validation
  */
 
-export const isErc20TransferCall = (call: Call): boolean => {
+export const isErc20TransferCall = (call: Call): call is Erc20TransferCall => {
   try {
     if (
       call &&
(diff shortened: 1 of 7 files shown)


packages/extension/src/shared/call.ts
```diff
@@ -28,13 +32,22 @@ export const isErc20TransferCall = (call: Call): boolean => {
...
       /** validate uint256 input amount */
-      if (isValidInputAmount(call.calldata[1])) {
-        return true
+      const amountUnit256: Uint256 = {
+        low: amountLowFelt,
+        high: amountHighFelt,
+      }
+      const amount = uint256ToBN(amountUnit256)
+      if (isUint256(amount)) {
+        if (toBN(amount).gt(toBN(0))) {
```
janek26, line 48: can be one if clause with &&

packages/extension/src/shared/call.ts
```diff
@@ -0,0 +1,81 @@
...
+  ]
+}
+
+/**
+ * Checks if the call has the expected shape of an ERC20 transfer call
+ * @param call Call
+ * @returns true if the call passes validation
+ */
+
+export const isErc20TransferCall = (call: Call): boolean => {
```
janek26, line 27:
```suggestion
export const isErc20TransferCall = (call: Call): call is Erc20TransferCall => {
```

janek26 (approved):
looks good now :)

PR #765, opened 2022-06-15
fix: allow upgrade without fee on mainnet

janek26:
just possible until cairo v0.9.0 releases on mainnet


simonheys (approved):

janek26:
dont merge as fees just hit main net

PR #783, opened 2022-06-16
fix: identify account by address and network

janek26:


packages/extension/src/shared/wallet.service.ts
```diff
@@ -9,6 +9,7 @@ export const hasNewDerivationPath = (derivationPath?: string): boolean =>
   Boolean(derivationPath?.startsWith(baseDerivationPath))
 
 export const isDeprecated = ({ signer, network }: WalletAccount): boolean => {
+  console.log("isDeprecated", signer.derivationPath)
```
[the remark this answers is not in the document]
janek26: thanks

packages/extension/src/background/wallet.ts
```diff
@@ -162,45 +160,45 @@ export class Wallet {
...
   private async pushAccount(account: WalletAccount) {
-    const accounts = await this.getAccounts(true)
-    const index = accounts.findIndex((a) => a.address === account.address)
-    if (index === -1) {
-      accounts.push(account)
-    } else {
-      accounts[index] = account
-    }
-    return this.store.setItem("accounts", accounts)
+    return this.setAccounts([account])
```
delaaxe, line 173: why does the `pushAccount` method seem to reset all accounts except for the one pushed

packages/extension/src/shared/wallet.model.ts
```diff
@@ -9,7 +9,14 @@ export interface WithSigner {
   signer: WalletAccountSigner
 }
 
-export interface WalletAccount extends WithSigner {
+export interface UniqAccount {
```
delaaxe, line 12: it's not more unique than the other interface no? anyway avoid abbreviations

```suggestion
export interface BaseWalletAccount {
```

packages/extension/src/ui/features/accountTokens/tokens.state.ts
```diff
@@ -147,7 +147,11 @@ export const useTokensWithBalance = (): UseTokens => {
   )
 
   const { data, isValidating, error, mutate } = useSWR(
-    [selectedAccount?.address, "accountTokenBalances"],
+    [
+      selectedAccount?.address,
+      selectedAccount?.network?.id,
+      "accountTokenBalances",
+    ],
```
delaaxe, line 154: just for my understanding, does adding a param here reset the cache for next load?

packages/extension/src/ui/services/backgroundAccounts.ts
```diff
@@ -24,8 +24,11 @@ export const connectAccount = ({ address, network, signer }: Account) => {
   sendMessage({ type: "CONNECT_ACCOUNT", data: { address, network, signer } })
 }
 
-export const deleteAccount = async (address: string) => {
-  sendMessage({ type: "DELETE_ACCOUNT", data: address })
+export const deleteAccount = async (address: string, networkId: string) => {
+  sendMessage({
+    type: "DELETE_ACCOUNT",
+    data: { address, network: { id: networkId } },
```
delaaxe, line 30: does the networkId really need to be part of the Wallet object here, not as a separate param?

packages/extension/src/ui/services/backgroundAccounts.ts
```diff
@@ -39,8 +42,11 @@ export const deleteAccount = async (address: string) => {
   }
 }
 
-export const hideAccount = async (address: string) => {
-  sendMessage({ type: "HIDE_ACCOUNT", data: address })
+export const hideAccount = async (address: string, networkId: string) => {
+  sendMessage({
+    type: "HIDE_ACCOUNT",
+    data: { address, network: { id: networkId } },
```
delaaxe, line 48: idem

packages/extension/src/background/wallet.ts
```diff
@@ -83,10 +83,8 @@ function calculateContractAddress(
   ])
 }
 
-export const equalAccount = (
-  a: Pick<WalletAccount, "address" | "network">,
-  b: Pick<WalletAccount, "address" | "network">,
-) => a.address === b.address && a.network.id === b.network.id
+export const equalAccount = (a: UniqAccount, b: UniqAccount) =>
```
delaaxe, line 86: unrelated but this is better english

```suggestion
export const accountsEqual = (a: UniqAccount, b: UniqAccount) =>
```

packages/extension/src/shared/wallet.model.ts
```diff
@@ -9,7 +9,14 @@ export interface WithSigner {
...
+  address: string
+  network: {
+    id: string
+  }
+}
+
+export interface WalletAccount extends UniqAccount, WithSigner {
```
delaaxe, line 19: is this still needed? shouldn't all `WalletAccount`s be persisted with only id?

packages/extension/src/ui/features/accountTokens/tokens.state.ts
janek26: yes, but just for one load

packages/extension/src/ui/services/backgroundAccounts.ts
janek26: I just reused the `BaseWalletAccount` here
to change this we would need to transform this twice just to improve internal messaging

delaaxe (approved):

PR #785, opened 2022-06-16
feat: display dollar value

simonheys:
Uses the Argent API to fetch token and price data. This is ultimately exposed as some simple hooks which automatically invoke and refresh the API when mounted:

- `useTokenBalanceToCurrencyValue(token)` - when the token contains its balance
- `useTokenAmountToCurrencyValue(token, amount)` - when the token and amount are separate
- `useSumTokenBalancesToCurrencyValue(arrayofTokens)`

USD price fetching code has been removed from the background process as the raw fee token values are now converted in the UI using these hooks.

Adds a new dotenv var `REACT_APP_ARGENT_API_BASE_URL` - if this is missing then no error will result but no currency values will show. A warning will be logged to the console in development.

Screenshot 2022-06-17 at 15 38 44

Screenshot 2022-06-17 at 15 38 54

Screenshot 2022-06-17 at 16 00 54


packages/extension/.env.example
```diff
@@ -1 +1,3 @@
 SEGMENT_WRITE_KEY=
+# URL pointing to the API base without trailing slash e.g. https://foo.bar.com/v1
```
janek26, line 2: trailing slash should not matter, please use `url-join` when working with the value

packages/extension/src/ui/features/accountTokens/AccountSubheader.tsx
```diff
@@ -72,6 +85,11 @@ export const AccountSubHeader: FC<AccountSubheaderProps> = ({
           </AccountStatusText>
         )}
       </div>
+      {sumCurrencyValue !== undefined && (
```
janek26, line 88:
```suggestion
      {sumCurrencyValue && (
```
as it will be a string there's no reason of 0 being false

janek26 (approved):

packages/extension/.env.example
simonheys: Good point, updated PR to add that and a more explicit check, will no longer attempt to invoke API at all unless the env var is defined

janek26 (approved):
looks good!

PR #786, opened 2022-06-16
feat: add mulitcall token balance fetching

dhruvkelawala:


dhruvkelawala:
Should we also fetch Token Details using multicall? 

If yes, should it be done in a different PR?

janek26:
I dont think that we need that for now, as token just get fetched when we add them, right?
Would be a different PR/ticket anyways

packages/extension/src/ui/services/multicall.service.ts
```diff
@@ -0,0 +1,24 @@
...
+import { useSelectedAccount } from "./../features/accounts/accounts.state"
+import { useCurrentNetwork } from "../features/networks/useNetworks"
+
+const MULTICALL_ADDRESS: Record<string, string> = {
+  SN_MAIN: "0x0740a7a14618bb7e4688d10059bc42104d22c315bb647130630c77d3b6d3ee50",
+  SN_GOERLI:
+    "0x042a12c5a641619a6c58e623d5735273cdfb0e13df72c4bacb4e188892034bd6",
+}
+
+export function useMulticallContract(): Contract | null {
```
janek26, line 13: can we keep this function pure? and provide the account and network as arguments please?

packages/extension/src/ui/services/multicall.service.ts
```diff
@@ -0,0 +1,24 @@
+import { Abi, Contract } from "starknet"
+
+import MULTICALL_ABI from "../../abis/Mulitcall.json"
+import { useSelectedAccount } from "./../features/accounts/accounts.state"
+import { useCurrentNetwork } from "../features/networks/useNetworks"
+
+const MULTICALL_ADDRESS: Record<string, string> = {
```
janek26, line 7: this should be part of the Network type, so people can change the address they want to use
See `shared/Network.ts`

should also be a (optional) field when adding a token

packages/extension/src/ui/features/accountTokens/tokens.service.ts
```diff
@@ -107,6 +117,56 @@ export const fetchTokenBalance = async (
...
 
+export type BalancesMap = Record<string, BigNumber | undefined>
+
+export const useFetchAllTokenBalances = () => {
+  const multicallContract = useMulticallContract()
+
+  return useCallback(
+    async (tokenAddresses: string[], accountAddress: string) => {
+      if (!multicallContract) {
+        return
```
janek26, line 128: if there is no multi call contract we should fallback to the current behaviour of fetching each token balance individually.

janek26 (changes requested):
All the points I catched here were properly described in the jira ticket
Please make sure to read it once again

packages/extension/src/ui/services/multicall.service.ts
dhruvkelawala: Yeah, we can. But I have the habit of using hooks wherever possible. But okay.

packages/extension/src/ui/features/accountTokens/tokens.service.ts
dhruvkelawala: Yeah, sorry. Dunno why I marked the PR as ready. My bad.

packages/extension/src/ui/services/multicall.service.ts
janek26: do it the other way around :D build a habit having as many pure functions as possible
They can be shared with pure TS without React, like the background process.

And once you have all the business logic in pure TS and you need a hook around it, just build one around the already existing function

janek26:
If we have a proper feedback I would not add that to the dapp api (yet)

janek26:
make sure to update tests by running `yarn test -u` locally and push the updates! you need devnet running

packages/extension/src/ui/features/accountTokens/tokens.service.ts
```diff
@@ -107,6 +117,66 @@ export const fetchTokenBalance = async (
...
+        [tokenAddresses[i]]: balance,
+      }
+    }, {})
+  }
+
+  const compiledCalldata = stark.compileCalldata({
+    address: account.address,
+  })
+
+  const calls = tokenAddresses.flatMap((address) => [
```
janek26, line 150: can we abstract all the multi call relevant logic into a function inside `packages/extension/src/ui/services/multicall.service.ts` ? so it can be used in multiple places in the future.

packages/extension/src/ui/features/accountTokens/tokens.service.ts
```diff
@@ -107,6 +117,66 @@ export const fetchTokenBalance = async (
...
+    hash.getSelectorFromName("balanceOf"),
+    compiledCalldata.length,
+    ...compiledCalldata,
+  ])
+
+  const response = await multicallContract.aggregate(calls)
+
+  const results: string[] = response.result.map((res: any) => number.toHex(res))
+
+  const resultsIterator = results.flat()[Symbol.iterator]()
```
janek26, line 161: normally we dont work with iterators, should be easy to do with normal array operations

packages/extension/src/ui/features/accountTokens/tokens.service.ts
janek26: as the fallback is also quite generic, we could move that into that function aswell, that way we could simply use the multi call everywhere

dhruvkelawala: We discussed this and came to conclusion that due to the complexity involved with multicall functionality, it's better to do it in a different PR.

packages/extension/src/ui/features/accountTokens/tokens.service.ts
```diff
@@ -158,21 +146,43 @@ export const fetchAllTokenBalances = async (
 
   const results: string[] = response.result.map((res: any) => number.toHex(res))
 
-  const resultsIterator = results.flat()[Symbol.iterator]()
+  return results.reduce<BalancesMap>((acc, result, i) => {
+    if (i % 2 === 0) {
```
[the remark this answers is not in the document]
dhruvkelawala: Thanks, that's a great solution

packages/extension/src/ui/features/accountTokens/tokens.service.ts
```diff
@@ -107,6 +118,67 @@ export const fetchTokenBalance = async (
...
+export const fetchAllTokensBalance = async (
+  tokenAddresses: string[],
+  account: Account,
+  network: Network,
+) => {
+  const multicallContract = getMulticallContract(account, network)
+
+  if (!multicallContract) {
+    // if no multicall contract is found, fallback to Promises
+    return fetchTokensBalanceWithPromises(tokenAddresses, account)
```
janek26, line 132: should be named `fetchTokenBalancesWithoutMulticall` 
as both use promises and it makes more sense in general

janek26 (approved):

packages/extension/src/ui/features/accountTokens/tokens.service.ts
dhruvkelawala: Agreed.

PR #787, opened 2022-06-17
fix: identify account by network and address

janek26:


packages/extension/src/shared/wallet.model.ts
```diff
@@ -11,13 +11,13 @@ export interface WithSigner {
...
+  networkId: string
 }
 
 export interface WalletAccount extends BaseWalletAccount, WithSigner {
-  address: string
   network: Network
   hidden?: boolean
 }
+
+export const accountsEqual = (a: BaseWalletAccount, b: BaseWalletAccount) =>
```
delaaxe, line 22: should be in `.service` file. `.model` files are for types and interfaces only

packages/extension/src/ui/features/accounts/accounts.state.ts
```diff
@@ -1,11 +1,16 @@
+import find from "lodash-es/find"
```
delaaxe, line 1: lodash's tree shaking is pretty good so it should be enough to just import from "lodash-es"

packages/extension/src/ui/features/accounts/accounts.state.ts
```diff
@@ -32,7 +37,9 @@ export const useAccount = (address: string): Account | undefined =>
 
 export const useSelectedAccount = () =>
   useAccounts(({ accounts, selectedAccount }) =>
-    selectedAccount ? accounts[selectedAccount] : undefined,
+    selectedAccount
+      ? find(accounts, (x) => accountsEqual(x, selectedAccount))
```
delaaxe, line 41:
```suggestion
      ? find(accounts, (account) => accountsEqual(account, selectedAccount))
```

packages/extension/src/shared/wallet.model.ts
```diff
@@ -11,13 +11,13 @@ export interface WithSigner {
...
-    id: string
-  }
+  networkId: string
 }
 
 export interface WalletAccount extends BaseWalletAccount, WithSigner {
-  address: string
   network: Network
   hidden?: boolean
 }
```
delaaxe, line 20: WalletAccount has both a `network.id` and a `networkId`, not sure if intended

packages/extension/src/background/wallet.ts
```diff
@@ -156,8 +174,11 @@ export class Wallet {
     )
   }
 
-  private async addWalletAccounts(accounts: WalletAccount[]) {
-    const oldAccounts = await this.getAccounts(true)
+  private async addWalletAccounts(
+    accounts: WalletAccount[],
+    overwriteWith?: WalletAccount[],
```
delaaxe, line 179: "overrides" are often a code smell. if this is a one-off behavior it should better be handled at the call site. if not a one-off should better be in a separate method

packages/extension/src/background/wallet.ts
```diff
@@ -132,15 +133,32 @@ export class Wallet {
...
           if (!network) {
             throw new Error("Network not found")
           }
+
+          try {
+            if (account.network?.id) {
+              account.networkId = account.network.id
+              // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+              // @ts-ignore
+              delete account.network
```
delaaxe, line 155: cast to the appropriate type or create a new object?

packages/extension/src/shared/wallet.model.ts
janek26: it is, for now we dont differentiate between "StoredWalletAccount" and the interface used by the application, so we keep both

packages/extension/src/background/wallet.ts
janek26: fine for now, as we dont have a split between the stored type and the interface used by the extension

simonheys (approved):

PR #803, opened 2022-06-20
bugfix: clicking close on Set up account recovery screen always goes to Account 1 overview

simonheys:
feat: enable navigating 'back' from close icon
bugfix: clicking close on Set up account recovery screen always goes to Account 1 overview

https://user-images.githubusercontent.com/175607/174583210-1c3f847f-4985-48e5-8931-97429c1f9208.mov


packages/extension/src/ui/components/IconBar.tsx
```diff
@@ -35,7 +36,7 @@ export const IconBar: FC<IconBarProps> = ({ back, close }) => (
     )}
     <hr />
     {close && (
-      <Link to={isString(close) ? close : routes.accountTokens()}>
+      <Link to={close !== undefined ? (close as any) : routes.accountTokens()}>
```
dhruvkelawala, line 39:
```suggestion
      <Link to={isString(close) || close === -1 ? (close as any) : routes.accountTokens()}>
```
We shouldn't check for undefined here because we already checking it on line 38.

dhruvkelawala (changes requested):

janek26:
this does not fix the issue if you close the flow in a later screen, ie
it even breaks it (as not every screen has a route I guess.
In general I dont think `x` (close) should ever refer to -1 (back)

https://user-images.githubusercontent.com/6831124/174587371-da3123c0-8854-4d98-bcc5-d89625643211.mp4

:

simonheys:
Changed approach to capture a modal `returnTo` path in the query string. Feels clean and appropriate to store this piece of page-related state in the url, but let me know what you think. Downside is this gets passed to each relevant page.

https://user-images.githubusercontent.com/175607/174616407-a637bfa7-e1a8-4d18-944c-4d5d592dec90.mov

janek26 (approved):
I like this solution a lot!

dhruvkelawala (approved):

PR #804, opened 2022-06-20
fix: unique account items in list

janek26:


dhruvkelawala (commented):
The code looks good and should work but needs testing to identify the exact cause of issue

janek26:
@dhruvkelawala I identified the issue, see last commit

simonheys (approved):
Tested a production build and seems to work as expected. Would be great to add some tests around this code at some point

PR #805, opened 2022-06-20
bugfix: broken fullscreen mode

simonheys:
bugfix: multiple global style override not working in production build - refactor to use single global style with props


janek26:
was this tested in Firefox aswell?

PR #807, opened 2022-06-21
feat: show fee estimation error

dhruvkelawala:


packages/extension/src/ui/features/actions/FeeEstimation.tsx
```diff
@@ -218,15 +244,55 @@ export const FeeEstimation: FC<FeeEstimationProps> = ({
...
     </FieldGroup>
   )
 }
+
+const getParsedError = (errorTxt: string) => {
+  const regex = new RegExp("Error in the called contract", "g")
+
+  const matchAll = [...errorTxt.matchAll(regex)]
+
+  return errorTxt.slice(matchAll[1].index)
```
janek26, line 297: what does this do

what if matchAll[1] is undefined?

please try catch and return errorTxt by default

dhruvkelawala: Yes sure, but I just saw the pattern of how Starknet returns errors and most of the time it has two "Error in the called contract". So I am quite sure it won't fail, but yes, try-catch is needed

packages/extension/src/background/transactions/transactionMessaging.ts
```diff
@@ -58,9 +58,14 @@ export const handleTransactionMessage: HandleMessage<
...
           },
         })
-      } catch (error) {
+      } catch (error: any) {
         console.error(error)
-        return sendToTabAndUi({ type: "ESTIMATE_TRANSACTION_FEE_REJ" })
+        return sendToTabAndUi({
+          type: "ESTIMATE_TRANSACTION_FEE_REJ",
+          data: {
+            error: error.message,
```
janek26, line 66: as we expect a string, lets make sure it is one in any case
```suggestion
            error: error?.message?.toString.?() ?? error?.toString.?() ?? "Unkown error",
```

janek26 (approved):

PR #814, opened 2022-06-22
fix: rules of hooks

janek26:


packages/extension/.eslintrc.js
```diff
@@ -17,10 +17,12 @@ module.exports = {
     ecmaVersion: "latest",
     sourceType: "module",
   },
-  plugins: ["react", "@typescript-eslint"],
+  plugins: ["react", "react-hooks", "@typescript-eslint"],
   rules: {
     "react/jsx-no-target-blank": "off",
     "react/react-in-jsx-scope": "off",
+    "react-hooks/rules-of-hooks": "error", // Checks rules of Hooks
+    "react-hooks/exhaustive-deps": "off", // Checks effect dependencies
```
simonheys, line 25: What's the reason for this to be "off" and not the recommended "warn"?

janek26: ci would fail as warn is also not allowed
In general I think we can enable it in the future

but I'm not a big fan of this setting, as I tent to add less into the dependency array than recommended to reduce renders when possible, or to get a specific behaviour

packages/extension/src/ui/App.tsx
```diff
@@ -1,5 +1,5 @@
 import { ThemeProvider, createTheme } from "@mui/material"
-import { FC, Suspense } from "react"
+import { FC, Suspense, useEffect } from "react"
```
janek26, line 2: use effect never used

simonheys (approved):

PR #818, opened 2022-06-23
fix: account total balance

simonheys:
fix: pass `account` argument so that tokens balances are fetched


janek26 (approved):
sorry for breaking this

PR #820, opened 2022-06-23
feat: add storybook package

simonheys:
Adds a siloed package with [Storybook](https://storybook.js.org/) UI component explorer.

Small refactor in the extension package to separate the theme so it can be cleanly re-used inside Storybook.

`yarn storybook` from root will open the two basic example stories with components imported from `@argent-x/extension`

Productivity boost: developers can edit the components and they will be hot-reloaded in the Storybook while you work on them in isolation. Designers can review all possible versions and states of a real component in isolation.

In future, storybook can build a static website that could be deployed internally as a component library.


packages/extension/src/ui/App.tsx
```diff
@@ -10,58 +8,7 @@ import { ErrorBoundary } from "./components/ErrorBoundary"
...
-/** @see `./theme.d.ts` for adding additional variables to the theme */
-const theme = createTheme({
-  palette: {
-    mode: "dark",
-  },
-  margin: {
-    extensionInTab: "10%",
-  },
-})
+import { GlobalStyle, theme } from "./theme"
```
simonheys, line 11: These were split into a separate 'theme' file so they can be cleanly shared with Storybook package

janek26 (approved):
I like this a lot! 
in the long term it should probably be part of some `ui` package that can get shared

PR #836, opened 2022-06-27
feat: add copy seed phrase functionality

dhruvkelawala:


janek26 (approved):
should this only be available when opened from the settings page?
It does not show when going through the banner flow

simonheys:
Works from Settings page, but I also expected to see it when following the 'Set up account recovery' flow. Is there a good reason for the two differ?

dhruvkelawala:
Now we have CopySeedPhrase on both screens @simonheys @janek26

simonheys (approved):

PR #844, opened 2022-06-29
build: performance improvements

janek26:
uses esbuild to speed up (dev) builds by 3x-5x

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+yarn tsc --noEmit -p packages/extension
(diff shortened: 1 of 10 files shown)


simonheys:
Definitely a big speed improvement, also looks like this fixes source maps in dev for me which would be an amazing productivity boost.

I am seeing an error when trying to use the extension, possible user error, here are my steps to reproduce;

* Check out this branch
* Install dependencies `yarn`
* Make sure to remove current extension from browser and remove `dist` folder
* `yarn build` or `yarn dev` to create extension in usual `dist` folder
* Usual 'load unpacked' step to add extension to browser
* Click extension icon

Expected result
* Extension window appears with start screen

Actual result
* Tiny white square window, no user interface

Clicking 'inspect' on the tiny square gives me one of the two attached screenshots depending on whether it was `yarn build` or `yarn dev`. In both cases the message includes `ReferenceError: React is not defined`

yarn build error

yarn dev error

simonheys (approved):

dhruvkelawala:
Works fine

PR #845, opened 2022-06-29
feat: rework token list for updated design variants

simonheys:
Rework token list for updated design variants, honouring user settings. This replaces the dependency `numeral` with a custom function plus test suites that give the desired output for displaying currency and token values. Question over whether this formatting should match the mobile apps exactly, but it is tweakable.

Stories have been added to Storybook to show the updated components.

Screenshot 2022-06-29 at 14 33 47

Screenshot 2022-06-29 at 14 34 02

Screenshot 2022-06-29 at 12 34 48

Screenshot 2022-06-29 at 12 34 55

Screenshot 2022-06-29 at 12 35 00


janek26 (approved):
love to see the stories!
and the test coverage is also out of this world 🫶

Approved!

PR #846, opened 2022-06-29
feat: add max transfer feature with feeEstimation

dhruvkelawala:


dhruvkelawala:
Still in draft. Need to figure out various cases to handle:

1. Subtract maxFee from MAX amount only in case of ETH
2. How to handle negative values. i.e. Clicking on MAX when balance is less than maxFee. Currently, it sets a negative value obtained from `balance - maxFee` and disables `Send` button. Should this be handled more graciously?
3. Should we hide MAX button once it's clicked? i.e. `inputAmount === (balance - maxFee)` and show again when maxFee is updated and it's different from the previous maxFee? We can use [usePrevious()](https://usehooks.com/usePrevious/#:~:text=useHooks(%F0%9F%90%A0)&text=Hooks%20are%20a%20feature%20in,them%20in%20your%20next%20project.) hook for that.

justingruss:
2. Don't think it's a good UX to show a negative value. Instead if user clicks 'MAX' when insufficient funds to cover maxFee, should populate the field with the max value user has available but also display error state with messaging explaining why they cannot proceed. Send button should be disabled

3. In updated UI the 'MAX' button will be hidden and fiat value will be shown in its place. Didn't account for changing maxFee. Maybe we could update input value automatically? On current UI we can keep 'MAX' button visible after clicking

packages/extension/src/ui/features/accountTokens/TokenScreen.tsx
```diff
@@ -111,6 +151,64 @@ const SendSchema: Schema<SendInput> = object().required().shape({
...
+    throw new Error("Account, TokenAddress and Balance are required")
+  }
+
+  const { compileCalldata, estimatedFeeToMaxFee: addOverheadToFee } = stark
+
+  const call: Call = {
+    contractAddress: tokenAddress ?? "",
+    entrypoint: "transfer",
+    calldata: compileCalldata({
+      recipient: account?.address ?? "0",
```
janek26, line 173: same

packages/extension/src/ui/features/accountTokens/TokenScreen.tsx
```diff
@@ -111,6 +151,64 @@ const SendSchema: Schema<SendInput> = object().required().shape({
...
+  } = useSWR(
+    [
+      "maxEthTransferEstimate",
+      Math.floor(Date.now() / 60e3),
+      account.networkId,
+    ],
+    () => getEstimatedFee(call),
+    {
+      suspense: false,
+      refreshInterval: 15 * 1000,
```
janek26, line 191:
```suggestion
      refreshInterval: 15e3,
```

packages/extension/src/ui/features/accountTokens/TokenScreen.tsx
```diff
@@ -111,16 +151,79 @@ const SendSchema: Schema<SendInput> = object().required().shape({
...
+export const useMaxFeeEstimateForTransfer = (
+  tokenAddress?: string,
+  balance?: BigNumber,
+  account?: Account,
+): {
+  maxFee?: BigNumber
+  error?: any
+  loading: boolean
+} => {
+  if (!account || !balance || !tokenAddress) {
```
janek26, line 163: check if token is the fee token, otherwise you can return a 0

packages/extension/src/ui/features/accountTokens/TokenScreen.tsx
```diff
@@ -129,17 +232,65 @@ export const TokenScreen: FC = () => {
...
+  const handleMaxClick = async () => {
+    setMaxClicked(true)
+    setMaxInputAmount(token, maxFee)
+  }
+
+  const disableSubmit =
+    isSubmitting ||
+    (submitCount > 0 && !isDirty) ||
+    !inputAmount ||
+    !inputRecipient ||
```
janek26, line 290: I would remove those as that gets handled by validation

packages/extension/src/ui/features/accountTokens/TokenScreen.tsx
```diff
@@ -129,17 +232,65 @@ export const TokenScreen: FC = () => {
...
+    inputAmount > balance ||
+    parseFloat(inputAmount) < 0
```
janek26, line 292: should be part of validation (yup schema)

packages/extension/src/ui/features/accountTokens/TokenScreen.tsx
```diff
@@ -175,21 +326,37 @@ export const TokenScreen: FC = () => {
             navigate(routes.accountTokens())
           })}
         >
-          <ControlledInputText
+          <StyledControlledInput
             autoComplete="off"
-            control={control}
+            control={control as any}
```
janek26, line 331: why is that needed?

dhruvkelawala: I was adding better typings to `ControlledInputTextAlt` but didn't work. Maybe we should add typings to it.

packages/extension/src/ui/features/accountTokens/TokenScreen.tsx
```diff
@@ -129,17 +248,60 @@ export const TokenScreen: FC = () => {
...
 
+  const {
+    maxFee,
+    error: maxFeeError,
+    loading: maxFeeLoading,
+  } = useMaxFeeEstimateForTransfer(token?.address, token?.balance, account)
+
+  useEffect(() => {
+    if (maxClicked && maxFee && token) {
+      setMaxInputAmount(token, maxFee)
```
janek26, line 267: should be reset at some point
maybe reset if the user edits the value the next time

or just disable input after max click, with some option to reenable it while reseting this value

dhruvkelawala: Hmm, makes sense

dhruvkelawala: Done

packages/extension/src/ui/features/accountTokens/TokenScreen.tsx
```diff
@@ -129,17 +249,67 @@ export const TokenScreen: FC = () => {
...
+    if (token.balance && maxFee) {
+      const balanceBn = token.balance
+
+      const maxAmount = balanceBn.sub(maxFee.toString())
+
+      const formattedMaxAmount = utils.formatUnits(
+        maxAmount.toString(),
+        tokenDecimals,
+      )
+      setValue("amount", maxAmount.lte(0) ? tokenBalance : formattedMaxAmount)
```
janek26, line 288: you can use BN.max with your value and `0`

and why would we input the tokenbalance in case the user can not send anyways? we should probably block him here already

janek26 (commented):
one more thought:
what happens if I as a User that want to send as much as possible click the "Max button" and realize you dont put all of my balance in the field. If I now manually put my token balance in the value field, would it work? I think we should block people from forwarding to the next screen in case `token balance + fee > input amount`

dhruvkelawala:
> f I now manually put my token balance in the value field

Not allowed case should be `updatedInputAmount + fee > balance`

packages/extension/src/ui/features/accountTokens/TokenScreen.tsx
dhruvkelawala: We are blocking it with an error message as suggested by @justingruss in point 1

dhruvkelawala:
Ready to merge

packages/extension/src/ui/components/InputText.tsx
```diff
@@ -118,6 +187,61 @@ export const ControlledInputText = styled(
...
+          style={{ position: "relative" }}
+          {...props}
+          value={value || ""}
+          {...field}
+          inputRef={ref}
+          inputMode="decimal"
+          type="text"
+          onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
+            console.log(
+              "🚀 ~ file: InputText.tsx ~ line 218 ~ e",
```
janek26, line 210: console.log


and let me know which viscose plugin this is

janek26 (approved):
looks good!

packages/extension/src/ui/components/InputText.tsx
dhruvkelawala: [Tubro Console Log](https://marketplace.visualstudio.com/items?itemName=ChakrounAnas.turbo-console-log)

PR #847, opened 2022-06-29
test: e2e

janek26:


package.json
```diff
@@ -18,12 +18,14 @@
   ],
   "scripts": {
     "format": "prettier --loglevel warn --write \"**/*.{js,jsx,ts,tsx,css,md,yml,json}\"",
-    "posttest": "npm run format",
     "dev": "lerna run dev --parallel",
     "clean": "rm -rf packages/extension/dist packages/get-starket/dist",
     "build": "lerna run build",
     "build:sourcemaps": "GEN_SOURCE_MAPS=true lerna run build",
-    "prepare": "yarn --cwd packages/get-starknet build && husky install",
+    "test": "lerna run test",
```
simonheys, line 25: I had an issue running just `yarn test` from the project root - one of the tests in extension would fail (because I wasn't running the starknet-devnet), but also `vitest` seems to run by default in 'watch' mode when not in CI? This means the test run never finishes, and there is no feedback as the test messages are swallowed by lerna.

Two possible workarounds I can suggest, you may have a better idea;
a) Change to `"test": "lerna run test --stream",` - this will stream test messages into the users terminal, so you can see any messages and will know that `vitest` has entered 'watch' mode
b) Explicitly make `vitest` not run in watch mode unless it is specified by the user - not sure the cleanest way to do that. Maybe have separate `yarn test` and `yarn test:watch` at root and/or package level?

package.json
```diff
@@ -18,12 +18,14 @@
...
+    "test:watch": "yarn --cwd packages/extension run test",
```
simonheys, line 26: Ah! yes see my comment above - `vitest` is in 'watch' mode by default

packages/extension/e2e/welcome.test.ts
```diff
@@ -0,0 +1,13 @@
+import { expect } from "@playwright/test"
+
+import { test } from "./fixture"
+import { openExtension } from "./steps/open-extension"
+
+test("Welcome page is shown correctly", async ({ page, context }) => {
+  await openExtension(page, context)
+
+  await expect(page.locator("text=Get started")).toBeVisible() // page is live
```
simonheys, line 9: On first-ever 'from cold' run, this test fails presumably as the 'loading spinner' phase takes longer than the test timeout?

On a second run the tests pass.

Is there a way to make sure the tests run from cold every time, ie. with no previous extension data? Perhaps it's controllable by the user as it might get annoying while iterating on tests?

Screenshot 2022-06-30 at 10 07 14

package.json
janek26: changed this, should be good now

packages/extension/e2e/welcome.test.ts
janek26: extended it to 8s, but it's most likely due to the alternating of the different greetings

simonheys (approved):

packages/extension/e2e/welcome.test.ts
dhruvkelawala: Failed for me on the first run as well

janek26: I think I fixed it

PR #849, opened 2022-06-29
fix: enable currency display only when on mainnet

simonheys:
Adds a `useIsMainnet` hook and uses it to check if the user is both on mainnet, and has Argent services enabled, before fetching and displaying token / pricing data

Video shows a user initially on Testnet with only token values displayed, switches to Mainnet showing both token and currency values ($0.00...!), then disabling Argent services which causes only token values. to be shown on Mainnet.

https://user-images.githubusercontent.com/175607/176479547-3a587863-24b1-416e-af45-2501cfb73150.mov

### DIFF
--- a/packages/extension/src/ui/features/networks/useNetworks.ts
+++ b/packages/extension/src/ui/features/networks/useNetworks.ts
@@ -52,3 +52,9 @@ export const useCurrentNetwork = () => {
 
   return currentNetwork.network
 }
+
+export const useIsMainnet = () => {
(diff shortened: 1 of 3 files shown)


packages/extension/src/ui/features/networks/useNetworks.ts
```diff
@@ -1,3 +1,4 @@
+import { useMemo } from "react"
```
janek26, line 1: unused?

janek26 (approved):
looks good to me!

PR #851, opened 2022-06-30
Logo update

Rahul-2110:
Extension logo and main logo updated.


janek26:
why exactly? :D

janek26:
please don't steal our application! And if so, make sure to respect the [License](https://github.com/argentlabs/argent-x/blob/develop/License.md)!

PR #853, opened 2022-07-01
Feature/improve-anonymous-analytics

janek26:


packages/extension/src/ui/features/actions/ActionScreen.tsx
```diff
@@ -97,6 +98,9 @@ export const ActionScreen: FC = () => {
           transactions={action.payload.transactions}
           actionHash={action.meta.hash}
           onSubmit={async () => {
+            analytics.track("signedTransaction", {
+              networkId: account?.networkId || "unknown",
```
dhruvkelawala, line 102: Do you think we can also track the contract address of the transaction calls? It might give us a great idea that which dApp is making use of ArgentX the most for signing transactions which can be different from preAuthorized dapps. Might also help us in prioritising those dApps for Rich Transaction Review.

simonheys (approved):

packages/extension/src/ui/features/actions/ActionScreen.tsx
janek26: we already do something which captures all dapp connection approvals, which is also a very good measure on how popular dapps are.
I feel like tracking contract addresses could be too personal

PR #861, opened 2022-07-02
test(e2e): add "send max eth" test

janek26:
as an example to showcase how I imagine us using e2e tests, and which structure to use
Let me know your feedback

### DIFF
--- a/packages/extension/e2e/selectors/getAccountAddressFromAccountPage.ts
+++ b/packages/extension/e2e/selectors/getAccountAddressFromAccountPage.ts
@@ -1,0 +1,19 @@
+import { expect } from "@playwright/test"
+import type { Page } from "@playwright/test"
+
+export async function getAccountAddressFromAccountPage(
+  page: Page,
+): Promise<string> {
+  await page.click("text=Add funds")
+  await page.click("text=From another StarkNet account")
+  const fullAddressEl = await page.waitForSelector(
+    "[aria-label='Full account address']",
+  )
+  const addressText = await fullAddressEl.textContent()
+  if (!addressText) {
+    return expect(addressText).toBeTruthy() as never
+  }
+  const address = addressText.replaceAll(" ", "")
+  await page.click("[aria-label='Close']")
+  return address
+}
--- a/packages/extension/e2e/sendMaxTokens.test.ts
+++ b/packages/extension/e2e/sendMaxTokens.test.ts
@@ -1,0 +1,54 @@
+import { expect } from "@playwright/test"
+
+import { test } from "./fixture"
+import { getAccountAddressFromAccountPage } from "./selectors/getAccountAddressFromAccountPage"
+import { getBalanceFromAccountPage } from "./selectors/getBalanceFromAccountPage"
+import { setupNewAccountWithTestnetEth } from "./setups/accountWithTestnetEth"
+import { approveTransaction } from "./steps/approveTransaction"
+import { navigateFromAccountToAccountList } from "./steps/navigateFromAccountToAccountList"
+import { navigateFromAccountToTokenDetails } from "./steps/navigateFromAccountToTokenDetails"
+import { newAccount } from "./steps/newAccount"
+import { selectAccountFromAccountList } from "./steps/selectAccountFromAccountList"
+import { waitForAllPendingTransactionsInAccount } from "./steps/waitForAllPendingTransactionsInAccount"
+import { formatTruncatedAddress } from "./utils"
+
+test("send max eth flow", async ({ page, context }) => {
+  const { address: a1, balance: b1 } = await setupNewAccountWithTestnetEth(
+    page,
+    context,
+  )
+  await navigateFromAccountToAccountList(page)
+  await newAccount(page)
+  const a2 = await getAccountAddressFromAccountPage(page)
+  const b2 = await getBalanceFromAccountPage(page, "Ether")
+  expect(b2).toBe("0.0")
+  await navigateFromAccountToAccountList(page)
+  await selectAccountFromAccountList(page, a1)
+  await navigateFromAccountToTokenDetails(page, "Ether")
+
+  await page.click("button:has-text('MAX')")
+  await page.fill("input[placeholder='Recipient']", a2)
+  await page.click("button:has-text('Send'):not([disabled])")
+
+  await approveTransaction(page)
+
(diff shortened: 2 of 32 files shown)


packages/extension/e2e/steps/openExtension.ts
```diff
@@ -1,6 +1,7 @@
 import type { BrowserContext, Page } from "@playwright/test"
 
 export async function openExtension(page: Page, context: BrowserContext) {
+  await page.waitForTimeout(1000) // give the extension 2s to startup
```
dhruvkelawala, line 4: 2s or 1s?

packages/extension/e2e/steps/newWallet.ts
```diff
@@ -0,0 +1,35 @@
...
+    degradedNetworkPerformanceText
+      .then(() => "degraded" as const)
+      .catch(() => "timeout" as const),
+  ])
+
+  if (whichWasFirst === "degraded") {
+    await page.click("I understand")
+  }
+
+  expect(await (await deployingText).isVisible()).toBe(true)
```
dhruvkelawala, line 34:
```suggestion
expect((await deployingText).isVisible()).resolves.toBe(true)
```

dhruvkelawala (commented):
Amazing! Writing these tests seem so much fun!

packages/extension/e2e/steps/openExtension.ts
janek26, line 4:
```suggestion
  await page.waitForTimeout(1000) // give the extension time to startup
```

Readme.md
```diff
@@ -124,13 +126,13 @@ For convenience this service can be started with Docker, note if successful ther
 - For ARM computers (e.g. Mac computers with Apple silicon)
 
   ```bash
-  docker run -p 5050:5050 shardlabs/starknet-devnet:latest-arm
+  docker run -p 5050:5050 shardlabs/starknet-devnet:latest-arm --seed 0
```
simonheys, line 129:
```suggestion
  docker run -it -p 5050:5050 shardlabs/starknet-devnet:latest-arm --seed 0
```

Adding `-it` causes output to show in the console so you can see if it's working or now

Readme.md
```diff
@@ -124,13 +126,13 @@ For convenience this service can be started with Docker, note if successful ther
...
   ```
 
 - Otherwise
 
   ```bash
-  docker run -p 5050:5050 shardlabs/starknet-devnet:latest
+  docker run -p 5050:5050 shardlabs/starknet-devnet:latest --seed 0
```
simonheys, line 135:
```suggestion
  docker run -it -p 5050:5050 shardlabs/starknet-devnet:latest --seed 0
```

packages/extension/e2e/selectors/getAccountAddressFromAccountPage.ts
```diff
@@ -0,0 +1,19 @@
+import { expect } from "@playwright/test"
+import type { Page } from "@playwright/test"
+
+export async function getAccountAddressFromAccountPage(
+  page: Page,
+): Promise<string> {
+  await page.click("text=Add funds")
```
simonheys, line 7: Is it worth adding test id's and using `data-testid=...` rather than checking for specific text here, just to future-proof from future text changes?

packages/extension/e2e/sendMaxTokens.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  await selectAccountFromAccountList(page, a1)
+  await navigateFromAccountToTokenDetails(page, "Ether")
+
+  await page.click("button:has-text('MAX')")
+  await page.fill("input[placeholder='Recipient']", a2)
+  await page.click("button:has-text('Send'):not([disabled])")
+
+  await approveTransaction(page)
+
+  await page.waitForSelector("h3:has-text('Pending transactions')")
```
simonheys, line 35: If the heading markup changes from `h3` to something else, will this break unintentionally?

packages/extension/src/ui/features/accountTokens/TokenScreen.tsx
```diff
@@ -326,7 +326,9 @@ export const TokenScreen: FC = () => {
           <BalanceTitle>Your balance</BalanceTitle>
           <InlineBalanceAndSymbol>
             <CopyToClipboard text={balance}>
-              <BalanceAmount>{balance}</BalanceAmount>
+              <BalanceAmount data-testid="tokenBalance">
```
simonheys, line 329: Use of `data-testid=` looks good, I would be more comfortable using this as a more resilient selector for tests

simonheys:
This is absolute 🔥 when it works. Very powerful. I found an issue getting it to run:

Steps to reproduce
* Check out any other branch e.g. `feature/show-fee-when-upgrading-impl`
* Run `yarn` and `yarn dev` as if you were developing on the branch
* Switch back to this PR branch
* Run `yarn` and `yarn test:e2e`

Expected result
* Tests eventually pass

Actual result
* Tests fail every time

Notes
I found I was able to get the tests to pass every time if I deleted the build folder first `rm -fr packages/extension/dist` then `yarn test:e2e` works every time. So possible some kind of 'clean build' step is needed?

packages/extension/e2e/selectors/getAccountAddressFromAccountPage.ts
janek26: ohhh! We can have a discussion on that :D Til now nobody won against my proposed "selection order" from here:
https://testing-library.com/docs/queries/about#priority

test-id's should be the last resort imo

packages/extension/e2e/sendMaxTokens.test.ts
janek26: I would expect this text to break yes. Which imo is better than outside factors making the test break, eg:
If it would only be text, without h3 and I add another "Pending transactions can not shown be shown" error (<p>) to the page and it shows, this test would still pass

packages/extension/e2e/selectors/getAccountAddressFromAccountPage.ts
simonheys: That order is totally backwards! From now on you have to pay me 1ETH every time a text change breaks a test.

But seriously, this is understood and fine and I think we should just all agree to follow the same principle you linked, maybe even consider adding it to the README.

simonheys (approved):

PR #862, opened 2022-07-03
fix: dont cache initial action queue values

janek26:


dhruvkelawala:
What's the reason behind doing this?

janek26:
useSWRImmutable used to work without cache
It seems like there was an update that it uses the cache now

we dont want to cache these values *at all* as they contain sensitive data/get outdated very quickly

dhruvkelawala (approved):

PR #896, opened 2022-07-04
feat: move to Aspect official API

0xs34n:


packages/extension/src/ui/features/accountNfts/aspect.service.ts
```diff
@@ -3,15 +3,22 @@ import join from "url-join"
...
 import { AspectNft } from "./aspect.model"
 
-const baseUrl = "https://api-testnet.playoasisx.com/assets"
+export const baseUrl = "https://api-testnet.aspect.co/api/v0/assets"
 
 export const fetchAspectNfts = async (
   account: BaseWalletAccount,
+  url: string,
 ): Promise<AspectNft[]> => {
   if (account.networkId === "goerli-alpha") {
```
AspectArchitect, line 12: as we're going to test mainnet, do we want to remove this goerli check?

packages/extension/src/ui/features/accountNfts/aspect.service.ts
```diff
@@ -3,15 +3,22 @@ import join from "url-join"
 import { BaseWalletAccount } from "../../../shared/wallet.model"
 import { AspectNft } from "./aspect.model"
 
-const baseUrl = "https://api-testnet.playoasisx.com/assets"
+export const baseUrl = "https://api-testnet.aspect.co/api/v0/assets"
 
 export const fetchAspectNfts = async (
   account: BaseWalletAccount,
+  url: string,
```
AspectArchitect, line 10: is url escaped? 

we just found a issue on our backend where url is not escaped and an error was thrown when we try to fetch

packages/extension/src/ui/features/accountNfts/aspect.service.ts
```diff
@@ -3,15 +3,22 @@ import join from "url-join"
...
 ): Promise<AspectNft[]> => {
   if (account.networkId === "goerli-alpha") {
     const params = new URLSearchParams({ owner_address: account.address })
-    const response = await fetch(join(baseUrl, `?${params}`))
-    return await response.json()
+    const response = await fetch(join(url, `?${params}`, "&limit=50"))
```
AspectArchitect, line 14: If our api is down due to AWS Lambda being down, this fetch will throw an error and Argent will crash.
Do we want to prevent Argent from crashing?

packages/extension/src/ui/features/accountNfts/aspect.service.ts
```diff
@@ -3,15 +3,22 @@ import join from "url-join"
...
+    const data = await response.json()
```
AspectArchitect, line 15: if url is incorrect response will be null and response.json() will throw an error.
I think we need `if (response)` check here to prevent processing null response

packages/extension/src/ui/features/accountNfts/aspect.service.ts
0xs34n: I think we can figure that out when we have mainnet NFTs

packages/extension/src/ui/features/accountNfts/aspect.service.ts
0xs34n: There is a try catch in `packages/extension/src/ui/features/accountNfts/AccountNfts.tsx` line 118

packages/extension/src/ui/features/accountNfts/aspect.service.ts
AspectArchitect, line 14: I think fetch() doesn't take the string as the second param. it's always an object containing request configs.

janek26 (approved):
thank you both!

PR #897, opened 2022-07-05
ci: even faster builds  🚀

janek26:


packages/extension/webpack.config.js
```diff
@@ -84,7 +86,12 @@ module.exports = {
       Buffer: ["buffer", "Buffer"],
       React: "react",
     }),
-    new ESLintPlugin({ extensions: ["ts", "tsx"], fix: true }),
+
+    ...(!isProd // eslint should run before the build starts
+      ? [new ESLintPlugin({ extensions: ["ts", "tsx"], fix: true })]
+      : []),
```
simonheys, line 98: This conditionally adding a plugin using spread could be a little hard to 'read', what do you think of this alternative pattern I see used e.g. in react-scripts?

```
plugins: [
  fooPlugin,
  !isProd && new ESLintPlugin({ extensions: ["ts", "tsx"], fix: true }),
  barPlugin
].filter(Boolean)
```

janek26: cool!

PR #898, opened 2022-07-05
fix: close popup after dapp extension unlock

janek26:
when there are no other actions that need approval


janek26:
I forgot to remove the playwright config changes, but I think they can go in aswell

simonheys (approved):

PR #901, opened 2022-07-05
feat: update token / currency tests and features for production API 

simonheys:
* updates snapshots from staging API
* updates tests to use updated data shape
* adds tests to show expected behaviour with invalid or missing data
* removes `REACT_APP_` prefix from dotenv var


janek26 (approved):
looks good!

PR #904, opened 2022-07-06
fix: use consistent storage key for account deploy transaction

simonheys:
Fixes an issue where all accounts showed a status of 'deploying' for a split second as the deploy transaction was never removed on success.

Users will initially see the 'deploying' on first use, then the deploy transaction will be removed and 'deploying' will no longer show on subsequent launches.

Screen recording shows deploying accounts successfully, then repeatedly opening the extension and viewing the account list - 'deploying' no longer shown.

https://user-images.githubusercontent.com/175607/177519814-f9e6b1c8-9dd6-4727-a678-b9b68ba7ab9f.mov

### DIFF
--- a/packages/extension/src/ui/features/accounts/Account.ts
+++ b/packages/extension/src/ui/features/accounts/Account.ts
@@ -41,7 +41,7 @@ export class Account {
       this.provider,
     )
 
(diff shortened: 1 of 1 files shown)


janek26:
thanks! Sorry I missed that

janek26 (approved):

PR #920, opened 2022-07-07
feat: add headers to argent api requests

simonheys:
Adds functions and tests for adding the expected Argent headers to Argent API requests, including the currently selected network.

The core feature is a new hook `useArgentApiFetcher()` which returns an SWR-compatible `fetcher` which adds the expected headers to all requests, adjusting for network changes automatically.

The spec is that requests sent to Argent API services should include the following http headers:

`argent-version` - e.g. `4.1.7`
`argent-client` - `argent-x`
`argent-network` - e.g. `goerli` or `mainnet`

### DIFF
--- a/packages/extension/src/shared/api/constants.ts
+++ b/packages/extension/src/shared/api/constants.ts
@@ -1,0 +1,30 @@
+import { isString } from "lodash-es"
+import urlJoin from "url-join"
+
+export const ARGENT_API_BASE_URL = process.env.ARGENT_API_BASE_URL as string
+
+export const ARGENT_API_ENABLED =
+  isString(ARGENT_API_BASE_URL) && ARGENT_API_BASE_URL.length > 0
+
+export const ARGENT_API_TOKENS_PRICES_URL = ARGENT_API_ENABLED
+  ? urlJoin(ARGENT_API_BASE_URL, "tokens/prices?chain=starknet")
+  : undefined
+
+export const ARGENT_API_TOKENS_INFO_URL = ARGENT_API_ENABLED
+  ? urlJoin(ARGENT_API_BASE_URL, "tokens/info?chain=starknet")
+  : undefined
+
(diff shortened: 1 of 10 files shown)


packages/extension/src/shared/api/constants.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const ARGENT_API_TOKENS_PRICES_URL = ARGENT_API_ENABLED
+  ? urlJoin(ARGENT_API_BASE_URL, "tokens/prices?chain=starknet")
+  : undefined
+
+export const ARGENT_API_TOKENS_INFO_URL = ARGENT_API_ENABLED
+  ? urlJoin(ARGENT_API_BASE_URL, "tokens/info?chain=starknet")
+  : undefined
+
+export const ARGENT_TRANSACTION_REVIEW_API_BASE_URL = process.env
```
simonheys, line 17: Looks like this might be the same as `ARGENT_API_BASE_URL` in production, so can eventually remove it and replace with `ARGENT_API_BASE_URL`

packages/extension/src/shared/api/fetcher.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/** convert KnownNetworksType to 'goerli' or 'mainnet' expected by API */
+
+export const argentApiNetworkForNetwork = (network: KnownNetworksType) => {
+  return network === "goerli-alpha" ? "goerli" : "mainnet"
+}
+
+export const argentApiHeadersForNetwork = (network: KnownNetworksType) => {
+  return {
+    "argent-version": process.env.VERSION || "",
```
janek26, line 44:
```suggestion
    "argent-version": process.env.VERSION || "Unknown version",
```

janek26 (approved):
I really enjoy your PRs! nice to see those test on such a consistent basis

PR #927, opened 2022-07-08
fix: token list exception

simonheys:
Fixes (for me) issues with the token list vs error handling. 

Root cause was a mix of swr suspense, and an issue with caching an error as an empty object. The suspense implementation was using a possibly undefined key (not allowed in suspense[1]) and returning empty data with a cached error (also not allowed in suspense[1]).

This causes an unrecoverable swr suspense error - reloading the component retrieves the same error from the cache. The error gets written to the cache at some point when reloading the extension, so reloading does not resolve it either. I was not able to identify where the empty object originates as it happens while reloading the extension and the debugger is closed.

I chose to always show the first token 'ETH' while the others are still initially loading as it made the app feel more responsive, however this could be changed to hide all until fetched. Previously no loading state was shown as the first token resolved the suspense immediately.

Some zustand selectors were memoized and shallow compare added to reduce chance of re-renders.

Adds `yarn dev:ui` which adds permanent 'reload' and 'Reset cache' buttons to the extension window.

[1] https://github.com/vercel/swr/pull/357#issuecomment-627089889

https://user-images.githubusercontent.com/175607/178026879-c17e2f19-468e-48fb-a6cd-4d14ae63aaa3.mov

### DIFF
--- a/packages/extension/src/shared/utils/error.ts
+++ b/packages/extension/src/shared/utils/error.ts
@@ -1,7 +1,10 @@
 export const UNKNOWN_ERROR_MESSAGE = "Unknown error"
 
-export const coerceErrorToString = (error: any): string => {
-  const errorObject = getErrorObject(error)
+export const coerceErrorToString = (
+  error: any,
+  includeStack = true,
+): string => {
+  const errorObject = getErrorObject(error, includeStack)
   if (errorObject) {
     return JSON.stringify(errorObject, null, 2)
   }
@@ -10,12 +13,13 @@ export const coerceErrorToString = (error: any): string => {
 }
 
 /** Convert an Error to an object with keys and values by introspecting keys */
-export const getErrorObject = (error: any) => {
+/** Conditionally allow tests to exclude the stack trace which is environment-specific */
+
+export const getErrorObject = (error: any, includeStack = true) => {
   try {
     if (typeof error === "object" && error !== null) {
-      /** ignore the stack trace which is environment-specific and handled by ErrorBoundary */
-      const keys = Object.getOwnPropertyNames(error).filter(
-        (key) => key !== "stack",
+      const keys = Object.getOwnPropertyNames(error).filter((key) =>
+        includeStack ? true : key !== "stack",
       )
       const errorObject: Record<string, string> = {}
       keys.forEach((key) => {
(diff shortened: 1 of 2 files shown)


packages/extension/src/ui/features/accountTokens/AccountTokens.tsx
```diff
@@ -131,30 +131,31 @@ export const AccountTokens: FC<AccountTokensProps> = ({ account }) => {
       )}
       {showNoBalanceForUpgrade && <UpgradeBanner canNotPay />}
       <PendingTransactions account={account} />
-      <ErrorBoundary
```
simonheys, line 134: No Suspense, the error is passed back from the hook instead

packages/extension/src/ui/features/accountTokens/tokens.state.ts
```diff
@@ -114,41 +116,82 @@ export const useTokensWithBalance = (
...
+  const tokenAddressesSelector = useMemo(() => {
+    return (state: State) => {
+      const tokensInNetwork = tokensInNetworkSelector(state)
+      return tokensInNetwork.map((t) => t.address)
+    }
+  }, [tokensInNetworkSelector])
+
+  // shallow compare objects and arrays
+  const tokensInNetwork = useTokens(tokensInNetworkSelector, shallow)
+  const tokenAddresses = useTokens(tokenAddressesSelector, shallow)
```
simonheys, line 133: Mix of memoized selectors and shallow compare should prevent new-but-same-array-of-tokens from triggering re-render

packages/extension/src/ui/features/dev/DevUI.tsx
```diff
@@ -0,0 +1,46 @@
...
+    background-color: rgba(255, 255, 255, 0.25);
+  }
+`
+
+const DevUI: FC = () => {
+  const reset = useCallback(() => {
+    // reset cache
+    const backupState = useBackupRequired.getState()
+    localStorage.clear()
+    useBackupRequired.setState(backupState)
```
janek26, line 29: can we share this function with the error screen?

packages/extension/src/ui/App.tsx
```diff
@@ -24,6 +25,7 @@ export const App: FC = () => {
           rel="stylesheet"
         />
         <GlobalStyle extensionIsInTab={extensionIsInTab} />
+        {process.env.SHOW_DEV_UI && <DevUI />}
```
janek26, line 28: I like this

PR #929, opened 2022-07-11
feat: Define ArgentX theme and make it the single-source of truth

dhruvkelawala:


packages/extension/src/ui/theme/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+  css,
+} from "styled-components"
+import { normalize } from "styled-normalize"
+
+import { Colors } from "./styled"
+
+const white = "#FFFFFF"
+const black = "#000000"
+
+export const colors = (): Colors => {
```
simonheys, line 16: What's the reason for this to be a function? Expect to pass some parameters to it in future?

packages/extension/src/ui/theme/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+}
+
+const MEDIA_WIDTHS = {
+  sm: 600,
+  md: 900,
+  lg: 1200,
+  xl: 1536,
+}
+
+const mediaMaxWidthTemplates: {
```
simonheys, line 51: This is clever, maybe too clever for me so I also find it hard to reason about. Can you add comments at least that show the expected shape of the templates and how they would be used?

packages/extension/src/ui/theme/styled.d.ts
```diff
@@ -0,0 +1,56 @@
+import { FlattenSimpleInterpolation } from "styled-components"
+
+export type Color = string
+export interface Colors {
```
simonheys, line 4: Why is this replicating the shape of colors function? Could be using e.g. typeof colors if colors was not a function?

packages/extension/src/ui/theme/styled.d.ts
```diff
@@ -0,0 +1,56 @@
...
+  sm: ThemedCssFunction<DefaultTheme>
+  md: ThemedCssFunction<DefaultTheme>
+  lg: ThemedCssFunction<DefaultTheme>
+  xl: ThemedCssFunction<DefaultTheme>
+}
+
+declare module "styled-components" {
+  export interface DefaultTheme extends Colors {
+    // css snippets
+    flexColumnNoWrap: FlattenSimpleInterpolation
```
simonheys, line 47: Would be great to see more comments or stories around how these are expected to be used

packages/extension/src/ui/theme/styled.d.ts
dhruvkelawala: It's the other way around. This defines the structure/shape of `colors`. And I agree, Colors doesn't have to be function as of now. Might only be useful when we introduce lightMode/darkMode.

packages/extension/src/ui/theme/styled.d.ts
dhruvkelawala: Added better docs

packages/extension/src/ui/theme/index.tsx
dhruvkelawala: Added docs

packages/extension/src/ui/theme/styled.d.ts
```diff
@@ -0,0 +1,56 @@
+import { FlattenSimpleInterpolation } from "styled-components"
+
+export type Color = string
```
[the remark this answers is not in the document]
dhruvkelawala: I don't agree with you on this one. `colors` object should be derived from `Colors` type and not the other way around. That's what typings are for. I don't understand what's the point of doing it the other way around. We always make an object from a type and not type from an object.

simonheys (approved):

dhruvkelawala:
@simonheys Using `typeof` colors

PR #931, opened 2022-07-12
feat: add notification badge

dhruvkelawala:
Screenshot 2022-07-12 at 12 59 39 PM


packages/extension/src/background/transactions/onchain.ts
```diff
@@ -11,6 +12,12 @@ export async function getTransactionsUpdate(transactions: Transaction[]) {
     transactionStatusToCheck.includes(status),
   )
 
+  browserAction.setBadgeText({
+    text: transactionsToCheck.length ? String(transactionsToCheck.length) : "",
+  })
+
+  browserAction.setBadgeBackgroundColor({ color: "#29C5FF" })
```
janek26, line 19: please move this into the `TransactionUpdateListener` handler in https://github.com/argentlabs/argent-x/blob/develop/packages/extension/src/background/transactions/notifications.ts#L12

packages/extension/src/background/transactions/transactions.ts
```diff
@@ -95,6 +95,7 @@ export const getTransactionsTracker: GetTransactionsTracker = (
       ...transaction,
     }
     await transactionsStore.addItem(newTransaction)
+    await handleUpdate()
```
janek26, line 98: if you move it this is not required

janek26 (changes requested):

packages/extension/src/background/transactions/onchain.ts
dhruvkelawala: Yep. I thought about it as well. However, I thought as we already had pending transactions available in the above file, I should do it there

packages/extension/src/background/transactions/notifications.ts
```diff
@@ -45,4 +48,18 @@ export const trackTransations: TransactionUpdateListener = async (
       }
     }
   }
+
+  if (IS_BROWSER) {
```
janek26, line 52: why is this needed? This won't run in the browser

packages/extension/src/background/transactions/notifications.ts
```diff
@@ -9,16 +11,17 @@ import { TransactionUpdateListener } from "./transactions"
 
 const successStatuses = ["ACCEPTED_ON_L1", "ACCEPTED_ON_L2", "PENDING"]
 
-export const trackTransations: TransactionUpdateListener = async (
-  transactions,
+export const trackTransactions: TransactionUpdateListener = async (
+  newTransactions,
```
janek26, line 15: the first argument won't just be "new" transactions, but all transactions that updated compared to last time
so this should be `updatedTransactions`

packages/extension/src/background/transactions/notifications.ts
dhruvkelawala: If we don't add it, it will fail during unit tests. IS_BROWSER just checks if typeof window is not undefined.

packages/extension/src/background/transactions/notifications.ts
```diff
@@ -45,4 +48,18 @@ export const trackTransations: TransactionUpdateListener = async (
       }
     }
   }
+
+  if (IS_BROWSER) {
+    const pendingTransactions = [...newTransactions, ...allTransactions].filter(
```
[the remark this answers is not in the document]
dhruvkelawala: It won't have "all" transactions. It will have allTransactions except the newTransactions because of this order https://github.com/argentlabs/argent-x/blob/e89ff58869afff78f3eaa7a1732995667684c2d7/packages/extension/src/background/transactions/transactions.ts#L101

Now, I know I can call txStore.getItems after adding new transaction and that way I would get all transactions including newTransactions. I just thought it would be better to keep them separate but I am open to change the order and allTransactions can include newTransaction as well.

packages/extension/src/background/transactions/notifications.ts
dhruvkelawala: Cool

dhruvkelawala:
Closing this PR and rebasing this feature on #940

PR #932, opened 2022-07-12
Feature/new-storage-interface-for-tokens

janek26:
replace the token implementation from messaging system to the new storage class


packages/extension/test/setup.ts
```diff
@@ -10,3 +10,12 @@ try {
...
 }
+try {
+  ;(window as any).chrome = {
+    runtime: {
+      id: "test",
+    },
+  }
+} catch {
+  // do nothing
+}
```
simonheys, line 21: Found out about this yesterday

```
/** mock global fetch() */
vi.stubGlobal("fetch", fetch)

/** satisfy checks to allow use of `import browser from "webextension-polyfill"` */
vi.stubGlobal("chrome", { runtime: { id: "testid" } })
```

janek26: very nice!

make sure to also let me know your thoughts on the new storage implementations, as they would be used everywhere some day it's important we're all aligned

packages/extension/src/shared/storage/__test__/hooks.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+        </div>
+      )
+    }
+    render(<Component />)
+
+    await screen.findByText("Hello World!")
+    fireEvent.click(screen.getByText("change value"))
+    await screen.findByText("Hello World! See you tomorrow!")
+  })
+})
```
simonheys, line 103: Loving the tests! When running tests, all passing but this file shows a warning:

```
yarn test src/shared/storage/__test__/hooks.test.tsx
```

```
Warning: An update to Component inside a test was not wrapped in act(...).

When testing, code that causes React state updates should be wrapped into act(...):

act(() => {
  /* fire events that update state */
});
/* assert on the output */

This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
    at Component (/Applications/MAMP/htdocs/argent-x/packages/extension/src/shared/storage/__test__/hooks.test.tsx:100:42)
```

packages/extension/.eslintrc.js
```diff
@@ -27,7 +27,7 @@ module.exports = {
     "react/jsx-no-target-blank": "off",
     "react/react-in-jsx-scope": "off",
     "react-hooks/rules-of-hooks": "error", // Checks rules of Hooks
-    "react-hooks/exhaustive-deps": "warn", // Checks effect dependencies
+    "react-hooks/exhaustive-deps": "off", // I dont see the reason for this check, I find myself always trying to use as few dependencies as possible, so it makes no sense to check if all are included in the array
```
simonheys, line 30: LOL I guess one other option here is to use `eslint-disable` to set it 'off' where explicitly intended with comments in the code? I would expect 'off' to be the exception rather than the norm?

packages/extension/src/ui/features/accountTokens/tokens.state.ts
```diff
@@ -157,37 +75,10 @@ export const useTokensWithBalance = (
...
-      if (JSON.stringify(maybeEmptyObjectError) === "{}") {
-        console.warn("FIXME: Ignoring empty object {} error")
-        return
-      }
-    } catch (e) {
-      // ignore any stringify errors
-    }
-    return maybeEmptyObjectError
-  }, [maybeEmptyObjectError])
-
```
simonheys, line 190: This is all still needed unfortunately, I can still see this empty error propagating into the UI

simonheys:
This is excellent work! Tests give great confidence that all works as intended.

One bonus thing that I would find very useful with something this fundamental is comments, maybe a short overview of the various storage types and simple examples of how they should be used? That could be work done as a separate story later though.

Once merged I would like to try using this for the dapp connections, I think that will help flush out any further comments.

There is one part where a fix should be reinstated, added a comment.

simonheys (changes requested):
There is one part where a fix should be reinstated, added a comment.

packages/extension/.eslintrc.js
janek26: okay we can do this aswell!

packages/extension/src/ui/features/accountTokens/tokens.state.ts
janek26: I think I fixed the underlying issue now

packages/extension/src/ui/features/accountTokens/SendTokenScreen.tsx
```diff
@@ -206,14 +204,8 @@ export const SendTokenScreen: FC = () => {
...
-  }, [
-    maxClicked,
-    maxFee,
-    setMaxInputAmount,
-    token,
-    token?.address,
-    token?.networkId,
-  ])
+    // dont add token as dependency as the reference can change
+  }, [maxClicked, maxFee, setMaxInputAmount, token?.address, token?.networkId]) // eslint-disable-line react-hooks/exhaustive-deps
```
simonheys, line 208: Not a storage issue but is this issue fixed if `token` is memoized?

```
const token = useMemo(
  () => tokenDetails.find(({ address }) => address === tokenAddress),
  [tokenAddress, tokenDetails],
)
```

packages/extension/src/ui/features/accountTokens/TokenScreen.tsx
```diff
@@ -69,7 +69,7 @@ export const TokenScreen: FC = () => {
   const { tokenAddress } = useParams()
   const account = useSelectedAccount()
   const { tokenDetails } = useTokensWithBalance(account)
-
+  console.log(tokenDetails.map((t) => t.name))
   const token = tokenDetails.find(({ address }) => address === tokenAddress)
```
simonheys, line 73: Same comment about possibly memoizing this

packages/extension/src/ui/features/actions/LoadingScreen.tsx
```diff
@@ -26,7 +26,7 @@ export const LoadingScreen: FC = () => {
   const { progress, clearProgress } = useLoadingProgress()
 
   // reset to 'indeterminate' spinner type on unmount
-  useEffect(() => () => clearProgress(), [])
+  useEffect(() => () => clearProgress(), []) // eslint-disable-line react-hooks/exhaustive-deps
```
simonheys, line 29: Oof could fix this later but this one is because `useLoadingProgress` recreates `clearProgress` each time, probably needs converting to `useCallback` so its stable

packages/extension/src/ui/features/actions/transaction/useTransactionReview.ts
```diff
@@ -54,7 +54,8 @@ export const useTransactionReview = ({
       transactions,
       fetcher,
     })
-  }, [account, transactions])
+    // dont rerender when fetcher reference changes
+  }, [account, transactions]) // eslint-disable-line react-hooks/exhaustive-deps
```
simonheys, line 58: Hm this doesn't seem right, I will make a note to come back and look at it. These comments are really useful!

simonheys (approved):
Few tiny notes, but looks good to go! Looking forward to giving this a try.

packages/extension/src/ui/features/actions/LoadingScreen.tsx
janek26: I added a todo comment for this

packages/extension/src/ui/features/actions/transaction/useTransactionReview.ts
janek26: Added another TODO comment for this

PR #933, opened 2022-07-12
feat: add Sentry for Error Reporting

dhruvkelawala:


dhruvkelawala:
Basic Sentry added

janek26:
Hey, I think the goal never was to add general sentry reporting, but to allow the user to report errors *explicitly* on the error screen. 
So next to "Reload" and "copy error" there would be a button "report error"

This is due to the confidential nature of a wallet. We dont want to collect any data without the user submitting it

packages/extension/src/ui/services/sentry.ts
```diff
@@ -0,0 +1,28 @@
...
+import { useBackgroundSettingsValue } from "./useBackgroundSettingsValue"
+
+export const useSentryInit = () => {
+  const { value: automaticErrorReporting } = useBackgroundSettingsValue<
+    ISettingsStorage["privacyErrorReporting"]
+  >("privacyShareAnalyticsData")
+
+  useEffect(() => {
+    Sentry.init({
+      dsn: "https://c2c861914bb2448e873ea3235fe03e5e@o1315819.ingest.sentry.io/6567926",
```
simonheys, line 14: IMO this should be a `process.env.SENTRY_DSN` or similar as it might need to differ between environments

packages/extension/src/ui/services/backgroundSettings.ts
```diff
@@ -2,8 +2,11 @@ import { sendMessage, waitForMessage } from "../../shared/messages"
 import { SettingsStorageKey } from "../../shared/settings"
 
 export const getSetting = async (key: SettingsStorageKey) => {
-  sendMessage({ type: "GET_SETTING", data: { key } })
-  return waitForMessage("GET_SETTING_RES")
+  await sendMessage({ type: "GET_SETTING", data: { key } })
+  const { value } = await waitForMessage("GET_SETTING_RES", ({ data }) => {
+    return data.key === key
```
simonheys, line 7: Oof my bad for not doing this originally. Can the settings storage be migrated onto the new shared storage stack? If so we should add a jira ticket for it.

packages/extension/src/ui/components/ErrorBoundaryFallbackWithCopyError.tsx
```diff
@@ -94,24 +109,62 @@ ${displayStack}
...
+      })
+      Sentry.captureException(error)
+    })
+  }
+
+  const {
+    initialised: privacyErrorReportingInitialised,
+    value: privacyErrorReportingValue,
+    setValue: setPrivacyErrorReportingValue,
+  } = useBackgroundSettingsValue<ISettingsStorage["privacyErrorReporting"]>(
```
simonheys, line 125: Not keen on the repetition of `privacyErrorReporting` here

simonheys (commented):
Did you look into automated upload of the source maps to Sentry as part of this? Fairly essential. Could be a separate 'enhancement' ticket though, not sure how tricky it is to set up via github actions.

packages/extension/src/ui/components/ErrorBoundaryFallbackWithCopyError.tsx
```diff
@@ -94,24 +109,62 @@ ${displayStack}
     return fallbackErrorPayload
   }, [error, errorInfo])
 
+  const reportToSentry = () => {
```
simonheys, line 112: Won't make much difference here as the render is still impure but I'd like to see this as `useCallback(`

packages/extension/src/shared/settings.ts
```diff
@@ -1,11 +1,18 @@
...
+  privacyErrorReporting: false,
 }
 
 export type SettingsStorageKey = keyof ISettingsStorage
 
 export type SettingsStorageValue = ISettingsStorage[SettingsStorageKey]
 
-export const isPrivacySettingsEnabled =
-  (process.env.FEATURE_PRIVACY_SETTINGS || "false") === "true"
+export const isPrivacySettingsEnabled = true
```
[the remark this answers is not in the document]
janek26: should this be enabled?

janek26:
if you think you could do automatic source map uploads to sentry that would be great!

packages/extension/src/shared/settings.ts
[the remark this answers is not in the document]
dhruvkelawala: This has to be enabled for allowing user to change "Automatic Error Reporting" from settings > privacy

packages/extension/src/ui/services/backgroundSettings.ts
dhruvkelawala: Yes :) . I spent more than 2hours just figuring out a weird issue as this was not present. I wasn't sure what was creating the problem xD

packages/extension/src/ui/components/ErrorBoundaryFallbackWithCopyError.tsx
dhruvkelawala: As in the variable name or the setting entirely?

packages/extension/src/ui/services/sentry.ts
```diff
@@ -0,0 +1,28 @@
...
+      debug: process.env.NODE_ENV === "development",
+      environment: process.env.NODE_ENV,
+      integrations: (integrations) => {
+        if (automaticErrorReporting) {
+          return integrations
+        }
+
+        return integrations.filter(
```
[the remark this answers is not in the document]
dhruvkelawala: I think this is better as it just removes the part from integrations that automatically captures exceptions.

packages/extension/src/ui/components/ErrorBoundaryFallbackWithCopyError.tsx
dhruvkelawala: I don't see much value in that, but will do it

packages/extension/src/ui/services/backgroundSettings.ts
simonheys: Really sorry, I didn't know this could happen

packages/extension/src/ui/components/ErrorBoundaryFallbackWithCopyError.tsx
simonheys: Having to specify it in the type as well as in the function argument. Not a big deal.

simonheys:
> if you think you could do automatic source map uploads to sentry that would be great!

Sentry have cli tools for this, from memory it can be complicated to set up in CI, may be worth creating a separate Jira ticket and escalating if it becomes a pressing issue.

packages/extension/src/ui/components/ErrorBoundaryFallbackWithCopyError.tsx
```diff
@@ -94,24 +109,66 @@ ${displayStack}
...
   }, [error, errorInfo])
 
+  const reportToSentry = useCallback(() => {
+    Sentry.withScope((scope) => {
+      Object.keys(errorInfo).forEach((key) => {
+        scope.setExtra(key, errorInfo[key])
+      })
+      const eventId = Sentry.captureException(error)
+      console.log(
+        "🚀 ~ file: ErrorBoundaryFallbackWithCopyError.tsx ~ line 118 ~ Sentry.withScope ~ eventId",
```
janek26, line 119: log

packages/extension/src/ui/services/useBackgroundSettingsValue.ts
```diff
@@ -10,13 +10,15 @@ import { getSetting, removeSetting, setSetting } from "./backgroundSettings"
  * @param key the key of the value to use
  */
 
-export const useBackgroundSettingsValue = (key: keyof ISettingsStorage) => {
+export const useBackgroundSettingsValue = <T extends SettingsStorageValue>(
+  key: keyof ISettingsStorage,
```
janek26, line 14: please do:

```typescript
export const useBackgroundSettingsValue = <T extends ISettingsStorage, K extends keyof T>(
   key: K,
): T[K] => {
```

packages/extension/src/ui/components/ErrorBoundaryFallbackWithCopyError.tsx
janek26: please clean this up, makes no sense to declare the type 2 times :D 
I added a comment to the useBackgroundSettingsValue file on how to do that

packages/extension/webpack.config.js
```diff
@@ -32,11 +33,7 @@ module.exports = {
   performance: {
     hints: false,
   },
-  devtool: isProd
-    ? genSourceMaps
-      ? "hidden-source-map"
-      : undefined
-    : "inline-source-map",
+  devtool: "source-map",
```
janek26, line 36: can't we keep the above?

janek26 (approved):

janek26 (changes requested):
misclick

packages/extension/webpack.config.js
dhruvkelawala: Yep.. without `source-map`, Sentry can't use sourcemaps. It needs `sourceMappingUrl`. However, we can use "inline-source-map" for dev and disable Sentry logging for dev. I don't think so its needed anyways.

packages/extension/src/ui/components/ErrorBoundaryFallbackWithCopyError.tsx
dhruvkelawala: Have left this here temporarily. Will remove before final merge. `eventId` helps debugging with sentry-cli

packages/extension/src/ui/services/useBackgroundSettingsValue.ts
```diff
@@ -10,13 +10,15 @@ import { getSetting, removeSetting, setSetting } from "./backgroundSettings"
...
+) => {
```
simonheys, line 15: Had to play around a bit as this is good learning for me too, what do you think?

A few changes together seem to get it working nicely, including changing the signatures of getSetting() etc.;

This gives the correct 'storedValue' state type
```
export const useBackgroundSettingsValue = <K extends keyof ISettingsStorage>(
  key: K,
) => {
  const [initialised, setInitialised] = useState<boolean>(false)
  const [storedValue, setStoredValue] = useState<ISettingsStorage[K]>()
```

Then can use the type e.g. here
```
const setValue = useCallback(
    async (value: ISettingsStorage[K]) => {
```

Same pattern in backgroundSettings
```
export const getSetting = async <K extends keyof ISettingsStorage>(
  key: K,
): Promise<ISettingsStorage[K]> => {

export const setSetting = async <K extends keyof ISettingsStorage>(
  key: K,
  value: ISettingsStorage[K],
) => {

export const removeSetting = async <K extends keyof ISettingsStorage>(
  key: K,
) => {
```

Then can simplify, e.g. this will know the return type from the key
```
const value = await getSetting(key)
```

dhruvkelawala: I wasted a lot of my time on this yesterday and potentially came out with a solution which didn't work. And we are going to change it to new storage type. i'll probably do it in other PR. Janek also agreed to it.

dhruvkelawala: If you agree to it, please approve :)

simonheys: IMO current state of these types is not good, it was cleaner and easier to follow before. We spend more time reading than writing code so legibility is paramount and is definitely not time wasted.

```
} = useBackgroundSettingsValue<ISettingsStorage["privacyShareAnalyticsData"]>(
    "privacyShareAnalyticsData",
  )
```

...needs to be like this to be legible:

```
} = useBackgroundSettingsValue(
    "privacyShareAnalyticsData",
  )
```

Let's defer to @janek26 to decide whether this PR or another one very soon after.

dhruvkelawala: It's waste of time just because we are going to change to Object storage type. That's why. Nothing else. Otherwise I also don't like the implementation currently.

janek26:
closed in favor of https://github.com/argentlabs/argent-x/pull/993

PR #940, opened 2022-07-13
Move networks and transactions to new storage

janek26:
waiting for #932 and will be way less commits to review afterwards


dhruvkelawala (approved):
Looks good to me. I think @simonheys should review it as well.

simonheys (approved):
All working, looks good!

PR #958, opened 2022-07-14
feat: grant dapps access to individual accounts 

simonheys:
Adds the ability for users to select individual accounts to connect to dapps, and shows connected status of the current dapp in the account list. Retains previous access behaviour so when users upgrade they will not see a change in behaviour until they modify their access.


packages/extension/test/tokenPrice.test.ts
```diff
@@ -10,11 +10,11 @@ import {
...
-import mockApiPricesDataInvalid from "./__mocks__/argent-api-prices-invalid.mock.json"
-import mockApiPricesData from "./__mocks__/argent-api-prices.mock.json"
-import mockApiTokenDataInvalid from "./__mocks__/argent-api-tokens-invalid.mock.json"
-import mockApiTokenData from "./__mocks__/argent-api-tokens.mock.json"
-import mockTokensWithBalanceRaw from "./__mocks__/tokens-with-balance.mock.json"
+import mockApiPricesDataInvalid from "./__fixtures__/argent-api-prices-invalid.mock.json"
+import mockApiPricesData from "./__fixtures__/argent-api-prices.mock.json"
+import mockApiTokenDataInvalid from "./__fixtures__/argent-api-tokens-invalid.mock.json"
+import mockApiTokenData from "./__fixtures__/argent-api-tokens.mock.json"
+import mockTokensWithBalanceRaw from "./__fixtures__/tokens-with-balance.mock.json"
```
simonheys, line 17: Moved these as they are technically fixtures rather than mocks

packages/extension/src/background/openUi.ts
```diff
@@ -43,8 +45,13 @@ async function openPopup() {
     left = Math.max(screenX + (outerWidth - NOTIFICATION_WIDTH), 0)
   }
 
+  /** popup loses context of originating host so it is passed in query */
+  const windowLocationHost = await getWindowLocationHost()
+  const host = windowLocationHost ? encodeURIComponent(windowLocationHost) : ""
+  const url = `index.html?popup&host=${host}`
```
janek26, line 51: this should be communicated through the created action item rather than the url

packages/extension/src/shared/messages/PreAuthorisationMessage.ts
```diff
@@ -1,15 +1,18 @@
+import { IHostAndAccount, IPreAuthorizations } from "../preAuthorizations"
 import { WalletAccount } from "../wallet.model"
 
 export type PreAuthorisationMessage =
   | { type: "CONNECT_DAPP"; data: { host: string } }
   | { type: "CONNECT_DAPP_RES"; data: WalletAccount }
-  | { type: "PREAUTHORIZE"; data: string }
```
janek26, line 6: should be kept, maybe rename to `REQUEST_PREAUTHORIZATION`

this should be emitted by the in page script and the background script should add a action to the actionQueue to accept or reject this action
after the user decided on the action the actual whitelisting is done

packages/extension/src/ui/features/actions/ApproveTransactionScreen.tsx
```diff
@@ -74,6 +77,12 @@ export const ApproveTransactionScreen: FC<ApproveTransactionScreenProps> = ({
     return <Navigate to={routes.accounts()} />
   }
 
+  /** will be undefined while initialising */
+  if (isPreauthorized === false) {
```
janek26, line 81: this should not be possible, is it?

simonheys: Not as far as I know

packages/extension/src/background/openUi.ts
janek26: from the `actionQueue`

maybe I get this wrong, but what is this parameter used for?

simonheys: This is needed when you open the extension by clicking the icon, so we can show 'connected' status of the current window host in the account list. It feels weird without this feature. I guess there may be other ways to pass this parameter into the extension, and I am open to ideas.

dhruvkelawala:
Testing locally

dhruvkelawala:
Also need to resolve conflicts

dhruvkelawala:
Really great work! This was a must needed UX and looking at the code it seems very difficult to say the least. Amazing!

dhruvkelawala:
Looking at this PR, can we also add a functionality where a dApp can request `disconnect`. In other words, use `removePreAuthorisation` for the selectedAccount by making a call on `window.starknet.disconnect` or `window.starknet.removeAuth`. Not sure if it should be done in this PR though as it's a very big PR already.

simonheys:
> Looking at this PR, can we also add a functionality where a dApp can request `disconnect`. In other words, use `removePreAuthorisation` for the selectedAccount by making a call on `window.starknet.disconnect` or `window.starknet.removeAuth`. Not sure if it should be done in this PR though as it's a very big PR already.

This is a good idea, would also be nice to see if there is a way to have a dApp 'reconnect' when switching accounts (I don't know if it's possible), will add a story.

dhruvkelawala (approved):

dhruvkelawala:
Need to fix e2e tests

janek26:
I’ll do that tomorrow morning, it’s related to e2e failing on develop

janek26:
done in #988, please review
once merge you can rebase :)

PR #966, opened 2022-07-18
fix: storybook swc and polyfill extension storage

simonheys:
Update Storybook build uses swc, polyfill extension storage and update expected types

### DIFF
--- a/packages/extension/src/ui/services/background.ts
+++ b/packages/extension/src/ui/services/background.ts
@@ -22,7 +22,7 @@ export const resetPreAuthorizatinos = () => {
   sendMessage({ type: "RESET_PREAUTHORIZATIONS" })
 }
 
-if (process.env.NODE_ENV === "development") {
+if (process.env.NODE_ENV === "development" && messageStream) {
   messageStream.subscribe(([message]) => {
     console.log("Received message", message)
   })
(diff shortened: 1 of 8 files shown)


packages/extension/src/ui/services/background.ts
```diff
@@ -22,7 +22,7 @@ export const resetPreAuthorizatinos = () => {
   sendMessage({ type: "RESET_PREAUTHORIZATIONS" })
 }
 
-if (process.env.NODE_ENV === "development") {
+if (process.env.NODE_ENV === "development" && messageStream) {
```
simonheys, line 25: This is undefined in storybook as it's not an extension

packages/storybook/package.json
```diff
@@ -18,6 +18,7 @@
     "eslint-plugin-storybook": "^0.6.1",
     "react": "^18.0.0",
     "react-dom": "^18.0.0",
+    "storybook-addon-swc": "^1.1.7",
```
simonheys, line 21: Compiles with swc (otherwise compilation now fails)

packages/storybook/.storybook/decorators.tsx
```diff
@@ -7,6 +8,18 @@ import {
...
+/** polyfill browser extension storage  */
+global.chrome = {
+  ...global.chrome,
+  runtime: {
+    ...global.chrome.runtime,
+    id: "test",
+  },
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  storage: chromeStorageMock,
```
simonheys, line 20: ...otherwise this fails type check as it's not an exhaustive implementation

PR #967, opened 2022-07-19
feat(analytics): add session ended analytics event

simonheys:


packages/extension/src/shared/analytics.ts
```diff
@@ -166,3 +172,55 @@ export function getAnalytics(
...
+  lastUnlocked: number
+  lastSession: number
+  lastClosed: number
+}
+
+interface ActiveStore extends ActiveStoreValues {
+  update: (key: keyof ActiveStoreValues) => void
+}
+
+export const activeStore = create<ActiveStore>(
```
simonheys, line 187: Had to move into shared because this also has to be updated by background

packages/extension/src/shared/analytics.ts
```diff
@@ -166,3 +172,55 @@ export function getAnalytics(
...
+    },
+  ),
+)
+
+/*
+ * There is no usable 'close' event on an extension
+ *
+ * instead we open a message port to the extension and simply listen for it to be disconnected
+ * as a side-effect of the extension being closed
+ */
```
simonheys, line 207: Other workarounds are available but involve polling, and this seemed much cleaner to me

packages/extension/src/ui/services/analytics.ts
```diff
@@ -120,7 +108,16 @@ export function trackAddFundsService(
...
+    // as React in strict mode renders every component twice, this will be called 2x in development. This is not the case in production.
     sessionStartTracking()
     openedExtensionTodayTracking()
+    return () => {
+      /**
+       * NOTE: any code here may run in dev but will not be triggered in a production build
+       * and therefore cannot be used for 'extension closed' events
+       *
+       * @see initUiExtensionCloseListener and initBackgroundExtensionCloseListener for solution
+       */
```
simonheys, line 120: Made a note here as this is not intuitive at all. Had a working version in dev before discovering it didn't work at all in a prod build.

packages/extension/src/ui/services/analytics.ts
```diff
@@ -100,10 +81,17 @@ export function unlockedExtensionTracking() {
 }
 
 export function sessionStartTracking() {
+  console.log("sessionStartTracking")
```
dhruvkelawala, line 84: Should this console log be removed?

dhruvkelawala (approved):
Just one comment, other than that everything looks good

PR #989, opened 2022-07-21
Add logging of network requests made through the wallet

tolbrino:
**NOTE: Work in progress**

This PR adds a view which shows logs for each network requests made by the wallet. The idea behind has been inspired by the work on [D.E.R.P.](https://derp.hoprnet.org/). The log view can be used by users to track what requests dApps are making through their wallet, or developers may use it to debug work done on the wallet.

# User Flow

- open extension
- open account switch view (top-left button)
- open settings (top-left button)
- click "Open Network Log" list entry
- then keep the new tab open and use the wallet, watch segment requests being logged

# Open Issues

- [ ] Log general fetch requests
- [ ] Fix reset of logs upon re-opening extension
- [ ] Log RPC requests made through the wallet

# Contributors

cc @QYuQianchen 
cc @tolbrino


janek26:
Hey :) It would be really nice to have a feature like this in ArgentX
I think it could be implemented easier by using https://developer.chrome.com/docs/extensions/reference/webRequest/ and filter by request origin to be the extension.
This could be extended to include approved dapp hosts at some point.

tolbrino:
> Hey :) It would be really nice to have a feature like this in ArgentX
> I think it could be implemented easier by using https://developer.chrome.com/docs/extensions/reference/webRequest/ and filter by request origin to be the extension.
> This could be extended to include approved dapp hosts at some point.

Thanks for the hint. That looks certainly easier to do than trying to hook into every place where requests are being triggered. Will check it out.

dhruvkelawala:
@tolbrino Are you still working on this PR or should this be closed?

janek26:
closed due to inactivity

feel free to reopen

PR #991, opened 2022-07-21
fix: suppress 429 error in token list

simonheys:
Adds a special case to the token list to suppress frequent errors (429 only at the moment). When the error matches this code, swr will retry again using default exponential backoff and the error will not propagate to parent.

### DIFF
--- a/packages/extension/src/ui/features/accountTokens/tokens.state.ts
+++ b/packages/extension/src/ui/features/accountTokens/tokens.state.ts
@@ -40,6 +40,9 @@ export const useToken = (baseToken: BaseToken): Token | undefined => {
   return token
 }
 
+/** error codes to suppress - will not bubble error up to parent */
+const SUPPRESS_ERROR_STATUS = [429]
+
 export const useTokensWithBalance = (
   account?: BaseWalletAccount,
 ): UseTokens => {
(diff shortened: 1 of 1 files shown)


packages/extension/src/ui/features/accountTokens/tokens.state.ts
```diff
@@ -62,24 +65,33 @@ export const useTokensWithBalance = (
...
       refreshInterval: 30000,
-      shouldRetryOnError: false,
+      shouldRetryOnError: (error) => {
+        return SUPPRESS_ERROR_STATUS.includes(error?.status)
+      },
     },
   )
 
-  const tokenDetailsIsInitialising = !error && !data && isValidating
+  const tokenDetailsIsInitialising = !error && !data
```
simonheys, line 94: Loading state continues if there is an initial error

dhruvkelawala (approved):

packages/extension/src/ui/features/accountTokens/tokens.state.ts
```diff
@@ -62,24 +65,33 @@ export const useTokensWithBalance = (
...
-        selectedAccount,
-      )
-
-      return balances
     },
     {
       refreshInterval: 30000,
-      shouldRetryOnError: false,
+      shouldRetryOnError: (error) => {
+        return SUPPRESS_ERROR_STATUS.includes(error?.status)
```
janek26, line 89: I guess this never gets called? because of the fetcher not throwing due to line #81 anyways?

janek26 (approved):

packages/extension/src/ui/features/accountTokens/tokens.state.ts
simonheys: Yes you're absolutely right, I have reverted to an earlier implementation I had where the error is checked outside of swr. I verified this works as expected by adding a random 429 error using our fetcher implementation which will throw the error if response is not ok:

```
if (Math.random() < 0.5) {
  return await fetcher("http://httpstat.us/429?sleep=100")
}
```

The UI shows as loading until it receives some data, and will show an error for all errors that do not have status 429

PR #993, opened 2022-07-22
Feat/update settings storage

dhruvkelawala:
This PR aims to update the Settings Storage to use the new ObjectStorage type. It's a follow-up/improvement PR to be merged after #933


packages/extension/src/ui/services/useBackgroundSettingsValue.ts
```diff
@@ -1,22 +1,29 @@
...
 /**
  * Hook providing access to an indiviudal value from background settings
  *
  * @param key the key of the value to use
  */
 
-export const useBackgroundSettingsValue = (key: keyof ISettingsStorage) => {
+export const useBackgroundSettingsValue = <K extends SettingsStorageKey>(
+  key: K,
+) => {
```
simonheys, line 20: So much nicer!

Might be neat to use the same approach on `getSetting()` and `setSetting()` functions so they are better constrained types too - but not a biggie

packages/extension/src/shared/messages/SettingsMessage.ts
```diff
@@ -16,7 +16,7 @@ export type SettingsMessageDataKeyValueRemoved = {
 
 export type SettingsMessage =
   | { type: "GET_SETTING"; data: SettingsMessageDataKey }
-  | { type: "GET_SETTING_RES"; data: SettingsStorageValue }
+  | { type: "GET_SETTING_RES"; data: SettingsMessageDataKeyValue }
```
simonheys, line 19: I think this whole file and references to anything it defines can be removed now?

dhruvkelawala: Working on that

packages/extension/src/shared/settings/index.ts
```diff
@@ -0,0 +1,32 @@
...
+import { settingsStorage } from "./storage"
+import { SettingsStorageKey, SettingsStorageValue } from "./types"
+
+export const getAllSettings = async () => {
+  return await settingsStorage.get()
+}
+
+export const getSetting = async <K extends SettingsStorageKey>(
+  key: K,
+): Promise<SettingsStorageValue> => {
```
simonheys, line 11: Return can be constrained further to specific value type by using the key K to pick from ISettingsStorage: `Promise<ISettingsStorage[K]>`

packages/extension/src/shared/settings/index.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  return allSettings[key]
+}
+
+export const setSetting = async <
+  K extends SettingsStorageKey,
+  V extends SettingsStorageValue,
+>(
+  key: K,
+  value: V,
```
simonheys, line 22: Same here instead of defining V could use `ISettingsStorage[K]` to get constrained type

simonheys (approved):

packages/extension/src/ui/services/useBackgroundSettingsValue.ts
```diff
@@ -1,22 +1,29 @@
...
+} from "../../shared/settings"
 
 /**
  * Hook providing access to an indiviudal value from background settings
  *
  * @param key the key of the value to use
  */
 
-export const useBackgroundSettingsValue = (key: keyof ISettingsStorage) => {
+export const useBackgroundSettingsValue = <K extends SettingsStorageKey>(
```
janek26, line 18: please simply use `useObjectStorage` hook for this.

janek26 (commented):
rest looks good!

packages/extension/src/ui/services/useBackgroundSettingsValue.ts
```diff
@@ -38,26 +30,16 @@ export const useBackgroundSettingsValue = <K extends SettingsStorageKey>(
...
-      }
+    if (!initialised) {
+      setInitialised(true)
     }
-  }, [key, storedValue, updateStoredValue])
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [key])
 
   return {
     initialised,
```
janek26, line 41: this value has no purpose anymore, as it will update on first render to true

PR #995, opened 2022-07-25
fix: dollar value of user input

simonheys:
Adds extra functions that are better suited to taking user input value (e.g. '1.23') for conversion to currency. Open to ideas for better naming :)

Uses these functions to fix an issue in the send screen where the dollar value was not updating.

https://user-images.githubusercontent.com/175607/180770078-61a08c11-52cb-4df5-88e5-427b64e09f6a.mov

### DIFF
--- a/packages/extension/src/ui/features/accountTokens/SendTokenScreen.tsx
+++ b/packages/extension/src/ui/features/accountTokens/SendTokenScreen.tsx
@@ -6,6 +6,7 @@ import styled from "styled-components"
 import { Schema, object } from "yup"
 
 import { inputAmountSchema, parseAmount } from "../../../shared/token/amount"
+import { prettifyCurrencyValue } from "../../../shared/token/price"
 import { getFeeToken } from "../../../shared/token/utils"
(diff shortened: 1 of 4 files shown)


packages/extension/src/ui/features/accountTokens/SendTokenScreen.tsx
```diff
@@ -167,7 +168,7 @@ export const SendTokenScreen: FC = () => {
   const inputAmount = formValues.amount
 
   const token = tokenDetails.find(({ address }) => address === tokenAddress)
-  const currencyValue = useTokenBalanceToCurrencyValue(token)
+  const currencyValue = useTokenUnitAmountToCurrencyValue(token, inputAmount)
```
simonheys, line 171: Open to ideas for naming and differentiating these. Updated the 'tsdoc' comments to help differentiate them.

janek26 (approved):
looks good!

In the long term I would love to see some of those hooks replaced by normal promises, and plug them into swr as a last step.
This would result in much more reusable and readable code imho

PR #997, opened 2022-07-25
fix: restrict input decimal length to default 16

simonheys:
Adds function and tests for checking numeric input, default restricted to max 16 decimals (longer than this causes other libraries e.g. ethers to fail)

### DIFF
--- a/packages/extension/test/tokenPrice.test.ts
+++ b/packages/extension/test/tokenPrice.test.ts
@@ -237,6 +237,12 @@ describe("prettifyTokenAmount()", () => {
           decimals: 18,
         }),
       ).toEqual("123,456,789.0")
+      expect(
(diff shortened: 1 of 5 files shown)


packages/extension/test/tokenPrice.test.ts
```diff
@@ -237,6 +237,12 @@ describe("prettifyTokenAmount()", () => {
           decimals: 18,
         }),
       ).toEqual("123,456,789.0")
+      expect(
+        prettifyTokenAmount({
+          amount: "100",
+          decimals: 18,
+        }),
+      ).toEqual("0.0000000000000001")
```
simonheys, line 245: testing display of 16 decimal places

janek26 (approved):

PR #998, opened 2022-07-25
fix: user can remove non-readonly default network

simonheys:
Adjustments so that default custom networks are non-readonly and can therefore be removed

Add a 'reset' feature so users can restore defaults

We accept that users who upgrade will lose 'localhost' from their custom networks however they can restore it using the 'reset' feature

https://user-images.githubusercontent.com/175607/180965121-0ad70b9f-fd1d-42f7-a8bb-28791cf23e1a.mov

### DIFF
--- a/packages/extension/src/ui/features/settings/NetworkSettingsFormScreen.tsx
+++ b/packages/extension/src/ui/features/settings/NetworkSettingsFormScreen.tsx
@@ -1,6 +1,6 @@
 import { Collapse } from "@mui/material"
-import { FC, useCallback, useEffect, useMemo, useState } from "react"
-import { WatchObserver, useForm } from "react-hook-form"
+import { FC, useEffect, useMemo, useState } from "react"
+import { useForm } from "react-hook-form"
 import { useNavigate } from "react-router-dom"
 import styled from "styled-components"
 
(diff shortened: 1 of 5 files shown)


packages/extension/src/ui/features/settings/NetworkSettingsFormScreen.tsx
```diff
@@ -70,7 +70,7 @@ export const NetworkSettingsFormScreen: FC<NetworkSettingsFormScreenProps> = (
 
   useEffect(() => {
     const subscription = watch((value, { name, type }) => {
-      if (type === "change" && name === "name") {
+      if (props.mode === "add" && type === "change" && name === "name") {
         setValue("id", slugify(value.name || ""))
```
simonheys, line 74: Only update the id if it's a new network - otherwise editing and changing the name results in a new network being added

janek26: good catch

packages/extension/src/ui/features/settings/NetworkSettingsFormScreen.tsx
```diff
@@ -1,6 +1,6 @@
 import { Collapse } from "@mui/material"
-import { FC, useEffect, useMemo, useState } from "react"
-import { useForm } from "react-hook-form"
+import { FC, useCallback, useEffect, useMemo, useState } from "react"
+import { WatchObserver, useForm } from "react-hook-form"
```
janek26, line 3: unused imports?

janek26 (approved):

PR #1012, opened 2022-07-28
Feat/address book

dhruvkelawala:


packages/extension/src/ui/components/DeleteDialog.tsx
```diff
@@ -0,0 +1,56 @@
+import { FC } from "react"
+import styled from "styled-components"
+import { DialogContentText, H3 } from "../theme/Typography"
+import { Button } from "./Button"
+import { ColumnCenter } from "./Column"
+import { CustomDialog } from "./CustomDialog"
+import Row from "./Row"
+import { darken } from "polished"
```
simonheys, line 8: The app is already using the lib `colord` for this kind of thing (chosen simply because it is very small). We probably shouldn't have two different deps for the same feature - I think either replace colord with polished everywhere, or use colord here instead.

packages/extension/src/ui/components/DeleteDialog.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+const DeleteButtonGroup = styled(Row)`
+  margin-top: 8px;
+  gap: 16px;
+`
+const CancelButton = styled(Button)`
+  background-color: ${({ theme }) => theme.text3};
+`
+const DeleteButton = styled(Button)`
```
simonheys, line 26: Does button with `variant="danger"` (see Storybook) give the correct appearance here? If not can you add a ticket in Jira so we can be sure to add another variant later.

packages/extension/src/ui/components/InputText.tsx
```diff
@@ -270,3 +274,109 @@ export const TextArea = styled.textarea`
...
+    control={control}
+    defaultValue={defaultValue}
+    rules={rules}
+    render={({ field: { ref, value, onChange: onValueChange, ...field } }) => (
+      <InputTextArea
+        style={{ position: "relative" }}
+        value={value || ""}
+        inputRef={ref}
+        maxRows={maxRows}
+        onChange={(e: React.ChangeEvent<HTMLTextAreaElement>) => {
```
simonheys, line 345: There was a function and tests added for this here I think, check if this does what is needed here? It was part of some work to restrict the input decimal places.

https://github.com/argentlabs/argent-x/blob/9daaaa2d091c415f266ffb1f85c53acef69ab142/packages/extension/src/ui/components/InputText.tsx#L244-L252

packages/extension/src/ui/features/accountTokens/SendTokenScreen.tsx
```diff
@@ -193,14 +260,35 @@ export const SendTokenScreen: FC = () => {
...
+    [setValue],
   )
 
   const [addressBookOpen, setAddressBookOpen] = useState(false)
+  const ref = useRef<HTMLDivElement>(null)
+  useOnClickOutside(ref, () => setAddressBookOpen(false))
 
   const addressBook = useAddressBook(account?.networkId || currentNetworkId)
 
+  const validateStarknetAddress = useCallback(
```
simonheys, line 272: Some of these functions are re-used? Can they me moved into a shared utility file (I would also like to see some simple unit tests to prove that these work as intended, but not essential)

dhruvkelawala: I agree that code needs refactoring because too many things are re-used. I personally hate the idea of writing something twice, so we should discuss in this PR regarding the most optimal way of doing things.

packages/extension/src/ui/components/DeleteDialog.tsx
dhruvkelawala: Yes, can use that. Sorry I missed it

packages/extension/src/ui/features/settings/AddressbookAddOrEditScreen.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+  const handleAddOrEditContact = async (
+    addressBookContactNoId: AddressBookContactNoId,
+  ) => {
+    let savedContact: AddressBookContact | undefined
+
+    if (currentMode === "add") {
+      savedContact = await addAddressBookContact({
+        ...addressBookContactNoId,
+        id: uniqueId("addressbook-contact"),
```
simonheys, line 145: Is there any consequence if `uniqueId` will return an existing id? (I would usually use something like uuid.v4 for a unique id here, but not sure of exact use-case)

packages/extension/src/ui/components/DeleteDialog.tsx
dhruvkelawala: Will remove `polished`

packages/extension/src/ui/components/InputSelect.tsx
```diff
@@ -0,0 +1,188 @@
+import styled, { css } from "styled-components"
+import Select, {
```
simonheys, line 2: How did you find working with `react-select`? We used the Mui select component on another screen and I found it was 'hostile' to style, would consider refactoring that with react-select if you think it's better?

packages/extension/src/ui/features/settings/AddressbookAddOrEditScreen.tsx
dhruvkelawala: I just read that `uniqueId` returns might return same output when the script is re-ran, so it was a bad idea after all. Oops

packages/extension/src/ui/components/InputSelect.tsx
```diff
@@ -0,0 +1,188 @@
...
+    inputRef,
+    defaultValue,
+    options,
+    classNamePrefix,
+    ...props
+  }: {
+    inputRef: any
+    style: HTMLProps<HTMLDivElement>["style"]
+  } & InputSelectProps) => {
+    const idRef = useRef(randomString())
```
simonheys, line 45: Not sure of use-case but wouldn't React `useId()` work here?

packages/extension/src/ui/components/InputSelect.tsx
dhruvkelawala: Even `react-select` was tough to style, but I think as I have created a reusable component, we should continue using react-select throughout

packages/extension/src/ui/components/InputSelect.tsx
dhruvkelawala: Yes, it will work. This line was written by Janek/Axel which I never bothered to change

simonheys:
I would expect the select to be constrained and scroll here, rather than the whole screen! (still usable, just unexpected)

https://user-images.githubusercontent.com/175607/181736019-e9073662-76f8-45f9-96e8-34e1f632c391.mov

simonheys:
Red underline on address - need to disable spellcheck maybe?

Screenshot 2022-07-29 at 10 26 03

simonheys:
Search is currently case-sensitive? It should return results regardless of case

https://user-images.githubusercontent.com/175607/181737301-2c62fefe-aedb-45f9-bdb9-aed7498975fe.mov

packages/extension/src/ui/features/settings/AddressbookSettingsScreen.tsx
```diff
@@ -0,0 +1,181 @@
...
+              <InputBefore>
+                <SearchIcon />
+              </InputBefore>
+            </SearchBox>
+
+            <AutoColumn gap="12px">
+              {contactList.map((contact) => {
+                const networkName = networkNameFromNetworkId(contact.networkId)
+                return (
+                  <ContactWrapper
```
simonheys, line 148: There is an existing reusable component for this, `AccountListItem` which would need a small modification to optionally show the network name, but I think otherwise should be used here for consistency.

(`yarn storybook` and you should be able to see it here http://localhost:6006/?path=/story/accounts-accountlistitem--outline )

simonheys:
Weird gap between these fields gets bigger once 'Max' has loaded

https://user-images.githubusercontent.com/175607/181739115-e1782588-7369-40bb-a7ff-cd346859856a.mov

simonheys:
The 'goerli testnet' looks like it is clickable, but it is actually disabled here? What is this for?

Screenshot 2022-07-29 at 11 25 17

dhruvkelawala:
> The 'goerli testnet' looks like it is clickable, but it is actually disabled here? What is this for?
> 
> Screenshot 2022-07-29 at 11 25 17

Forgot to update the `disabled` styling in react-select. My bad. Planned to do it at the end but just forgot about it

simonheys:
Bonus: I think it would be super-nice in these screens if the grey icon at the top changed into the colourful circle with initials (e.g. via `AccountAvatar` and `getNetworkAccountImageUrl()`) as you type the name (throttle changes obviously) - what do you think?

image

image

packages/extension/src/ui/features/accountTokens/SendTokenScreen.tsx
janek26: I remember I did a yup schema for addresses once, I think we should use yup way more
for validating values, but also building forms

janek26: https://github.com/argentlabs/argent-x/blob/20066e4f88cf8346de0e599b8769e22e0aeb745c/packages/extension/src/ui/services/addresses.ts#L35

packages/extension/src/ui/components/DeleteDialog.tsx
simonheys: Don't apologise - I think the problem is it's far too easy to miss. I've added a note of some improvements we should consider as part of the UX review work

dhruvkelawala:
> Search is currently case-sensitive? It should return results regardless of case
> 
>  case-sensitive-search.mov

Resolved

dhruvkelawala:
> > The 'goerli testnet' looks like it is clickable, but it is actually disabled here? What is this for?
> > Screenshot 2022-07-29 at 11 25 17
> 
> Forgot to update the `disabled` styling in react-select. My bad. Planned to do it at the end but just forgot about it

Resolved

dhruvkelawala:
> I would expect the select to be constrained and scroll here, rather than the whole screen! (still usable, just unexpected)
> 
>  select-scroll.mov

Resolved

packages/extension/src/ui/features/accountTokens/SendTokenScreen.tsx
dhruvkelawala: `addressSchema.validate` and `addressSchema.isValid` returns `Promise<boolean>`. Should we use Promises here?

packages/extension/package.json
```diff
@@ -89,13 +90,16 @@
...
+    "react-select": "^5.4.0",
+    "react-textarea-autosize": "^8.3.4",
     "starknet": "^3.18.2",
     "starknet-390": "npm:starknet@3.9.0",
     "styled-components": "^5.3.5",
     "styled-normalize": "^8.0.7",
     "swr": "^1.3.0",
     "tracking-link": "^4.0.1",
     "url-join": "^5.0.0",
+    "uuid": "^8.3.2",
```
janek26, line 102: I usually use https://bundlephobia.com/package/nanoid@4.0.0 if there is no strong need for uuid, what do you think? :) 7x smaller and native types

packages/extension/src/shared/addressBook/storage.ts
```diff
@@ -0,0 +1,10 @@
+import { ArrayStorage } from "../storage"
+import { AddressBookContact } from "./type"
+
+export const equalContact = (a: AddressBookContact, b: AddressBookContact) =>
+  (a.address === b.address && a.networkId === b.networkId) || a.id === b.id
+
+export const addressBookStore = new ArrayStorage<AddressBookContact>([], {
```
janek26, line 7: you love to see it! let me know if we can improve something when working with ArrayStorage

packages/extension/src/ui/components/CustomBottomSheet.tsx
```diff
@@ -0,0 +1,89 @@
+import { FC, useEffect, useState } from "react"
```
janek26, line 1: Just rename it to `BottomSheet`?

packages/extension/src/ui/components/CustomDialog.tsx
```diff
@@ -0,0 +1,37 @@
+import Dialog from "@mui/material/Dialog"
```
janek26, line 1: love it!

packages/extension/src/ui/components/InputText.tsx
```diff
@@ -101,7 +105,7 @@ const InputAlt = styled.input`
   }
 `
 
-function randomString() {
+export function randomString() {
```
janek26, line 108: unused, no?
should be replaced by useId

packages/extension/src/ui/features/accountTokens/SendTokenScreen.tsx
janek26: there's addressSchema.isValidSync too

packages/extension/src/ui/features/accountNfts/SendNftScreen.tsx
```diff
@@ -72,32 +93,72 @@ export const SendNftScreen: FC = () => {
...
 
+  const formValues = watch()
+  const inputRecipient = formValues.recipient
+
+  const validateStarknetAddress = useCallback(
+    (addr: string) => addr.length > 62 && addr.length <= 66, // including 0x
+    [],
+  )
+
+  const validRecipientAddress = useMemo(
```
janek26, line 133: should rely on the form hook validation and not introduce a 2nd kind of validation

packages/extension/src/ui/features/accountNfts/SendNftScreen.tsx
```diff
@@ -136,38 +224,97 @@ export const SendNftScreen: FC = () => {
...
+                    style={{
+                      paddingRight: "50px",
+                      borderRadius: addressBookOpen ? "8px 8px 0 0" : "8px",
+                    }}
+                    onChange={(e) => {
+                      if (validateStarknetAddress(e.target.value)) {
+                        const account = addressBook.contacts.find((c) => {
+                          return (
+                            normalizeAddress(c.address) ===
+                            normalizeAddress(e.target.value)
```
janek26, line 273: pretty sure there's some utility function around to compare addresses by numeric value

packages/extension/src/ui/features/accounts/ProfilePicture.tsx
```diff
@@ -3,10 +3,11 @@ import styled, { css } from "styled-components"
 export interface ProfilePictureProps {
   disabled?: boolean
   small?: boolean
+  size?: number
 }
 
-const sizeAttribute = ({ small }: ProfilePictureProps) =>
-  small ? "24px" : "36px"
+const sizeAttribute = ({ small, size }: ProfilePictureProps) =>
+  size ? `${size}px` : small ? "24px" : "36px"
```
janek26, line 10: should we replace `small` by `size`? so it's the same everywhere

janek26 (commented):
looks good! nice work!

packages/extension/src/ui/features/accountNfts/SendNftScreen.tsx
dhruvkelawala: Can't find

packages/extension/src/ui/features/accounts/ProfilePicture.tsx
dhruvkelawala: You mean `size ? `${size}px: "24px"`

janek26: just make size mandatory and put it everywhere
or (if it's just 3 variants) lets add variants, like
```
interface ProfilePicture {
  ...
  variant: "lg" | "md" | "sm"
}
```

packages/extension/src/ui/features/accountNfts/SendNftScreen.tsx
janek26: ah maybe it was v3 branch

janek26 (approved):

PR #1016, opened 2022-07-29
fix: adjust polling behaviour

simonheys:
Adds adjustments and tests to the 'polling' behaviour for token and price endpoints.

New behaviour
* Default revalidate on mount (otherwise you may never see fresh prices)
* Dedupe requests to poll interval (otherwise 'mount' behaviour above causes multiple requests)

Previous behaviour
* Lack of 'dedupe' could cause multiple requests when remounting components, e.g. switching accounts

### DIFF
--- a/packages/extension/src/ui/services/swr.ts
+++ b/packages/extension/src/ui/services/swr.ts
@@ -38,6 +38,14 @@ const swrPersistedCache: Cache = {
   },
 }
 
+/** SWR config - keep default behaviour with refresh and dedepe for 'polling' behaviour */
+export const withPolling = (interval: number) => {
+  return {
+    refreshInterval: interval,
+    dedupingInterval: interval /** dedupe multiple requests */,
+  }
+}
+
 /** SWR fetcher used by useConditionallyEnabledSWR when disabled */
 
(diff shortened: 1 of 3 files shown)


packages/extension/src/ui/services/swr.ts
```diff
@@ -54,14 +63,18 @@ export function useConditionallyEnabledSWR<Data = any, Error = any>(
...
     config,
   )
-  /** revalidate when enabled changes */
+  /** reset the cache when disabled */
   useEffect(() => {
-    result.mutate()
+    if (!enabled) {
+      result.mutate()
+      cache.delete(key)
+    }
```
simonheys, line 77: Adjustments and tests updated, works better with polling behaviour

packages/extension/test/swr.test.tsx
```diff
@@ -26,28 +32,126 @@ describe("swr", () => {
...
+          <Component />
+          <Component />
+          <Component />
+          <Component />
+          <Component />
+          <Component />
+          <Component />
+          <Component />
+          <Component />
+        </>,
```
simonheys, line 131: Multiple components tests the 'dedupe' behaviour

janek26 (approved):

PR #1021, opened 2022-08-03
feat: transaction review

simonheys:
Adds special display of Jediswap 'swap' transaction review payload including token, value and slippage for source and destination. Falls back to a default display if there is no transaction review, or if the review does not contain 'swap' information. Refactors some fields to make them more easily reusable in future.

Screenshot 2022-08-03 at 16 34 11

Screenshot 2022-08-03 at 16 36 33

Screenshot 2022-08-03 at 16 36 54

Screenshot 2022-08-03 at 16 37 09

Screenshot 2022-08-03 at 16 51 43


janek26 (approved):
such a nice PR! well done

PR #1031, opened 2022-08-10
Feat/argent plugin accounts

dhruvkelawala:


packages/extension/src/ui/features/settings/NetworkSettingsFormScreen.tsx
```diff
@@ -168,7 +168,7 @@ export const NetworkSettingsFormScreen: FC<NetworkSettingsFormScreenProps> = (
               autoComplete="off"
               control={control}
               placeholder="Account class hash"
-              name="accountClassHash"
+              name="accountClassHash.argentAccount"
```
janek26, line 171: this works? nice

we should add a field to add a plugin account of your choice at some point too

janek26 (commented):
nice!

Next up registering the plugin system? :)

packages/extension/src/ui/features/settings/NetworkSettingsFormScreen.tsx
dhruvkelawala: Yeah it works.

Yes, we should. For now, keeping it simple.

janek26:
this needs a starknet.js release before it's ready for review

packages/dapp/src/components/TokenDapp.tsx
```diff
@@ -115,6 +128,57 @@ export const TokenDapp: FC = () => {
...
+      if (!sessionAccount) {
+        throw new Error("No open session")
+      }
+      const erc20Contract = new Contract(
+        Erc20Abi as Abi,
+        getErc20TokenAddress(network),
+        sessionAccount,
+      )
+
+      const result = erc20Contract.mint(
```
simonheys, line 169:
```suggestion
      const result = await erc20Contract.mint(
```

packages/extension/src/shared/account/details/fetchType.ts
```diff
@@ -0,0 +1,107 @@
...
+              [
+                call.contractAddress,
+                hash.getSelectorFromName(call.entrypoint),
+                0,
+              ] as const,
+          )
+
+          const response = await multicallContract.aggregate(multicallCall)
+
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
simonheys, line 75: Are there unused vars here? Maybe used to be

packages/extension/src/shared/settings/index.ts
```diff
@@ -1,4 +1,7 @@
 export const isPrivacySettingsEnabled =
   (process.env.FEATURE_PRIVACY_SETTINGS || "false") === "true"
 
+export const isExperimentalSettingsEnabled =
+  (process.env.FEATURE_EXPERIMENTAL_SETTINGS || "false") === "true"
```
simonheys, line 5: I think `FEATURE_EXPERIMENTAL_SETTINGS` needs to be added into `.env.example` too?

packages/extension/src/ui/services/useHover.ts
```diff
@@ -0,0 +1,26 @@
...
+        node.addEventListener("mouseover", handleMouseOver)
+        node.addEventListener("mouseout", handleMouseOut)
+        return () => {
+          node.removeEventListener("mouseover", handleMouseOver)
+          node.removeEventListener("mouseout", handleMouseOut)
+        }
+      }
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [ref.current], // Recall only if ref changes
```
simonheys: Not sure the intention here, but AFAIK ref or ref.current won't work as dependencies and won't trigger the effect again. There is a solution here https://medium.com/@teh_builder/ref-objects-inside-useeffect-hooks-eb7c15198780 - please let me know if this is wrong as it's a pattern I have used elsewhere

packages/extension/src/shared/settings/index.ts
janek26: good catch

packages/extension/src/ui/services/useHover.ts
janek26: file removed

packages/sessions/README.md
```diff
@@ -0,0 +1,147 @@
+# Sessions
+
+Sessions can be used to send transactions from a dapp on behalf of a user without requiring his confirmation with a wallet.
```
simonheys, line 3:
```suggestion
Sessions can be used to send transactions from a dapp on behalf of a user without requiring their confirmation with a wallet.
```

simonheys (approved):

PR #1078, opened 2022-08-19
fix: add token to wallet rpc message

dhruvkelawala:


packages/extension/src/background/tokenMessaging.ts
```diff
@@ -0,0 +1,32 @@
...
+import { HandleMessage, UnhandledMessage } from "./background"
+
+export const handleTokenMessaging: HandleMessage<TokenMessage> = async ({
+  msg,
+  background: { actionQueue },
+  sendToTabAndUi,
+}) => {
+  switch (msg.type) {
+    case "REQUEST_TOKEN": {
+      // TODO: Check if token already exists
```
janek26, line 11: todo?

dhruvkelawala: Sorry, it's done. We don't need it here anymore.

dhruvkelawala:
ArgentX 2022-08-24 10-15-18

janek26 (approved):

PR #1080, opened 2022-08-22
add Game of Life Token to default tokens

thecotne:
GoL-128px X 128px

### DIFF
--- a/packages/extension/src/assets/default-tokens.json
+++ b/packages/extension/src/assets/default-tokens.json
@@ -100,6 +100,13 @@
     "image": "https://dv3jj1unlp2jl.cloudfront.net/128/color/usdt.png"
   },
   {
+    "address": "0x06dc4bd1212e67fd05b456a34b24a060c45aad08ab95843c42af31f86c7bd093",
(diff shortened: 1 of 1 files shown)


dhruvkelawala:
Why should this be a default token? It serves no purpose as a default token. You can give a button that says "Add to Wallet" (or something similar) which on click calls
`window.starknet.request({ type: "wallet_watchAsset", params: { type: "ERC20",  options: { address: tokenAddress } }})`. This will ask user if he wants to add this token to his and if the user approves, this token will be shown in his wallet

robwalsh21:
> Why should this be a default token? It serves no purpose as a default token. You can give a button that says "Add to Wallet" (or something similar) which on click calls `window.starknet.request({ type: "wallet_watchAsset", params: { type: "ERC20", options: { address: tokenAddress } }})`. This will ask user if he wants to add this token to his and if the user approves, this token will be shown in his wallet

Missed this @dhruvkelawala .  That works for us.

janek26:
but very nice token icon!

PR #1112, opened 2022-08-26
feat: pending transactions badge

simonheys:
This shows the count of pending transactions of the currently selected account with a new badge in the 'Activity' tab and changes to the count logic in the extension icon (which previously showed a count of pending actions). The extension icon will update in the background.


janek26:
can you try to build this with manifest v3 ? 
```
MANIFEST_VERSION=v3 yarn dev
```

Lets make sure it works with both versions (seems like chrome api is different for badges )

janek26 (approved):
thanks a lot!

PR #1114, opened 2022-08-30
feat: nft collections

dhruvkelawala:


janek26 (approved):
nice work! looks and works as expected.

In general I dont know if collection view should always be the default, but will take this to Justin

PR #1108, opened 2022-08-25
feat: new transaction feed

simonheys:
Adds explorer transactions which are matched against a set of known dapps and transaction types to display a richer UI.

This is currently implemented as an enhancement to the existing activity features, and will fallback to that implementation if disabled.

The activity screen will display enriched transactions, or externally received transactions when they become available in the feed.

`ARGENT_EXPLORER_BASE_URL` is required in dotenv for this feature to be enabled.

Screenshot 2022-08-25 at 11 11 20

Screenshot 2022-08-25 at 11 11 36

Screenshot 2022-08-25 at 11 11 52

Screenshot 2022-08-25 at 11 12 05

Screenshot 2022-08-25 at 11 12 13

Screenshot 2022-08-25 at 11 12 25

Screenshot 2022-08-25 at 11 11 11

Screenshot 2022-08-25 at 11 09 27
Screenshot 2022-08-25 at 11 09 32
Screenshot 2022-08-25 at 11 09 38
Screenshot 2022-08-25 at 11 09 46
Screenshot 2022-08-25 at 11 09 53
Screenshot 2022-08-25 at 11 09 57
Screenshot 2022-08-25 at 11 10 01
Screenshot 2022-08-25 at 11 10 13
Screenshot 2022-08-25 at 11 10 22

Screenshot 2022-08-25 at 11 09 16
Screenshot 2022-08-25 at 11 09 23

PR #1135, opened 2022-09-02
feat: pending transactions UI

simonheys:
Adds an additional transformer which uses raw calldata to determine some transaction types. This allows use of the new UI, which has been renamed to be more generic and the old components removed.

This is stacked onto https://github.com/argentlabs/argent-x/pull/1108


simonheys:
Merged into https://github.com/argentlabs/argent-x/pull/1108 for sanity

PR #1146, opened 2022-09-06
fix: gracefully handle malformed nft

simonheys:
Adds extra checks and flexibility so the UI will more gracefully handle missing or malformed NFT data.


packages/extension/src/ui/features/accountNfts/aspect.service.ts
```diff
@@ -45,56 +46,86 @@ export const fetchAspectNftsByUrl = async (
...
+  }
+
+  return data.assets
 }
 
 export const fetchAspectCollection = async (
   account?: BaseWalletAccount,
   contractAddress?: string,
-): Promise<Collection | null> => {
-  try {
```
simonheys, line 68: Allow these functions to throw so an error state is captured by swr or the UI - otherwise Suspense with no error and no data stays in an indefinite loading state.

janek26 (approved):
nice! thanks for looking into this
code looks also good

PR #1150, opened 2022-09-06
update session with RequestSession

gregoryguillou:
There should actually 2 be differents object:
- `Signer` and `ec` come from starknet.js and provides the command create a session key and sign transactions
- `sessionRequest: SessionRequest` and `createSession` are defined in @argent/x-sessions and are used to generate the signedSession

### DIFF
--- a/packages/sessions/README.md
+++ b/packages/sessions/README.md
@@ -44,7 +44,7 @@ Next you need to come up with the permissions you would like for your session. Y
 This example session will allow the dapp to execute an example endpoint on an example contract without asking the user to approve the transaction again. After signing the session the dapp can execute all transactions listed in `policies` whenever it wants and as many times as it wants.
 
 ```typescript
-import { Signer, ec } from "@argent/x-sessions"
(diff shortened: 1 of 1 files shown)


packages/sessions/README.md
```diff
@@ -44,12 +44,12 @@ Next you need to come up with the permissions you would like for your session. Y
 This example session will allow the dapp to execute an example endpoint on an example contract without asking the user to approve the transaction again. After signing the session the dapp can execute all transactions listed in `policies` whenever it wants and as many times as it wants.
 
 ```typescript
-import { Signer, ec } from "starknet"
+import { Signer, ec } from "@argent/x-sessions"
```
janek26, line 47: this comes from `starknet` still, right?
```suggestion
import { Signer, ec } from "starknet"
```

janek26 (changes requested):
rest looks good to me, thanks for the feedback!

janek26 (approved):

PR #1152, opened 2022-09-07
export the Policy interface

gregoryguillou:
It could be beneficial you export `Policy` especially because (1) you reference it in the doc and export requestSession and (2) you might want to change it and it would help if I do not redefine it. Obviously this configuration might also be on purpose to force using `preparePolicy`


janek26 (approved):
happy to export the type - as long as only high level implementations get exported

PR #1153, opened 2022-09-07
fix: notification trigger

simonheys:
Triggers notification on known transaction status change instead of load. This fixes an issue where restoring a wallet would trigger a notification for every single transaction.

### DIFF
--- a/packages/extension/src/background/notification.ts
+++ b/packages/extension/src/background/notification.ts
@@ -1,7 +1,22 @@
 import { Status } from "starknet"
 import browser from "webextension-polyfill"
 
+import { ArrayStorage } from "../shared/storage"
 import { TransactionMeta } from "../shared/transactions"
+
+const notificationsStorage = new ArrayStorage<string>(
+  [],
+  "core:notifications:seenTransactions",
+)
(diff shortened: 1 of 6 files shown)


janek26:
Ha, this opens the question if we want to notify users about transactions that complete which were made with their account using an external tool (tx fetched from voyager for the first time), which we do currently

packages/extension/src/background/notification.ts
```diff
@@ -1,23 +1,8 @@
 import { Status } from "starknet"
 import browser from "webextension-polyfill"
 
-import { ArrayStorage } from "../shared/storage"
 import { TransactionMeta } from "../shared/transactions"
 
-const notificationsStorage = new ArrayStorage<string>(
```
janek26, line 7: afaik this is still needed, as transactions can update from ie
PENDING -> ACCEPTED_ON_L2

as we show the push notification as soon as the transaction enters pending, we dont want to notify again when it enters ACCEPTED_ON_L2

This can also be triggered by fetching voyager

dhruvkelawala (approved):

PR #1171, opened 2022-09-08
fix: address field crash

simonheys:
Adds optional restriction to input that only allow user to type or paste a `0x0abc...` address. Adds additional changes and tests for address function(s) which were throwing on invalid input.


dhruvkelawala (approved):
Good one

PR #1172, opened 2022-09-08
fix: unable to add network

simonheys:
Adds deep form object value clean before validation, updated schema for required value inside optional object, adds tests


dhruvkelawala (approved):

janek26 (approved):
thank you!

PR #1175, opened 2022-09-09
fix: zero balance send

simonheys:
Refactors the dialog into a more generic component with Storybook. Uses this to display an actionable message to a new user when they click Send and their account is still deploying or has zero balance. Navigates to 'add funds' screen when clicked.

Screenshot 2022-09-09 at 16 53 43

Screenshot 2022-09-09 at 16 55 16

Screenshot 2022-09-09 at 16 55 39

Screenshot 2022-09-09 at 17 00 37

Screenshot 2022-09-09 at 17 00 41

Screenshot 2022-09-09 at 17 00 45

Screenshot 2022-09-09 at 17 00 49


janek26 (approved):
good refactor 🙏 no clue why it wasn't dont like this in the first place

PR #1181, opened 2022-09-14
fix: e2e tests

simonheys:
Updates to fix an issue with clicking MAX not enabling the 'Next' button, and slow network conditions allowing a stale balance to be captured in tests.

### DIFF
--- a/packages/extension/src/ui/features/accountTokens/TokenScreen.tsx
+++ b/packages/extension/src/ui/features/accountTokens/TokenScreen.tsx
@@ -15,6 +15,7 @@ import { routes } from "../../routes"
 import { H2, H3 } from "../../theme/Typography"
 import { useSelectedAccount } from "../accounts/accounts.state"
 import { TokenIcon } from "./TokenIcon"
+import { IIsLoading, isLoadingPulse } from "./TokenListItem"
 import { TokenMenu } from "./TokenMenu"
 import { useTokenBalanceToCurrencyValue } from "./tokenPriceHooks"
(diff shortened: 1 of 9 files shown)


packages/extension/src/ui/features/accountTokens/TokenScreen.tsx
```diff
@@ -95,16 +109,16 @@ export const TokenScreen: FC = () => {
...
-            <RowCentered
-              gap="8px"
-              align="baseline"
-              style={{ marginTop: "12px" }}
-            >
-              <H2 style={{ marginBottom: "0px" }} data-testid="tokenBalance">
+            <TokenBalanceContainer isLoading={isLoading}>
+              <StyledH2
+                data-testid={
+                  isLoading ? "tokenBalanceIsLoading" : "tokenBalance"
```
simonheys, line 115: Ensures the tests won't capture a stale value while it is still loading

packages/extension/src/ui/features/accountTokens/useMaxFeeForTransfer.ts
```diff
@@ -14,7 +14,7 @@ export const useMaxFeeEstimateForTransfer = (
   balance?: BigNumber,
   account?: Account,
 ): {
-  maxFee?: BigNumber
+  maxFee?: string
```
simonheys, line 17: Keeps value as a string which is stable when used as a dependency

janek26 (approved):

PR #1215, opened 2022-09-15
fix(ui): always allow delete in input fields

simonheys:
Adds a check for deletion so user can recover from invalid value. Adds stories.


janek26 (approved):
looks good :)

PR #1217, opened 2022-09-15
fix(ui): check and ignore sequencer transfers

simonheys:
Adds a check for unexpected events that are gas fee transfers to sequencer.


janek26:
shouldn't this be done by the BE ?

simonheys:
> shouldn't this be done by the BE ?

Ideally yes.

janek26:
merging this anyways, as it would work with or without the BE removing that entry
So we can ship a quick fix

janek26 (approved):

PR #1234, opened 2022-09-27
feature: full screen onboarding

simonheys:
Refactors and modifies the initial onboarding flow to use full screen


dhruvkelawala:
Is this PR ready for review?

packages/extension/src/ui/features/onboarding/ui/OnboardingButton.tsx
```diff
@@ -0,0 +1,9 @@
+import styled from "styled-components"
```
dhruvkelawala, line 1: Should this be in a separate file? I think it can be in the same file. wdyt?

packages/extension/src/ui/routes.ts
```diff
@@ -35,11 +35,14 @@ export const useReturnTo = () => {
...
-  seedRecovery: route("/recover/seed"),
-  seedRecoveryPassword: route("/recover/seed/password"),
+  onboardingStart: route("/index.html"),
+  onboardingDisclaimer: route("/onboarding/disclaimer"),
+  onboardingPrivacy: route("/onboarding/privacy"),
+  onboardingPassword: route("/onboarding/password"),
+  onboardingFinish: route("/onboarding/finish"),
+  onboardingRestoreBackup: route("/onboarding/restore/backup"),
+  onboardingRestoreSeed: route("/onboarding/restore/seed"),
+  onboardingRestorePassword: route("/onboarding/restore/password"),
```
dhruvkelawala, line 45: Better idea would be refactor them into a different object `onboardingRoutes` and then merge them here using spread

dhruvkelawala (changes requested):

dhruvkelawala:
Other than these minor changes, things look good

packages/extension/src/ui/features/browser/tabs.ts
```diff
@@ -20,7 +20,7 @@ export const focusExtensionTab = async () => {
 }
 
 export const useExtensionIsInTab = () => {
-  const [isInTab, setIsInTab] = useState(false)
```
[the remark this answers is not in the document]
simonheys: Yes because the initialisation is async so it is `undefined` when the state is 'not yet known'.

packages/extension/src/ui/features/onboarding/ui/OnboardingButton.tsx
simonheys: I think the buttons need a proper holistic refactor as part of the wider design system changes. This is pretty unique to Onboarding right now and I think better not to pollute the more general Button class with it.

packages/extension/src/ui/routes.ts
simonheys: Good point I agree there are better ways to declare the routing overall. However this is consistent with current approach for now, and I don't think we should introduce a new way of doing it as part of this PR.

packages/extension/src/ui/features/browser/tabs.ts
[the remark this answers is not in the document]
simonheys: `false` implies 'definitely not in a tab', but as this is async initially we don't know if we are in a tab or not and should wait for a definite `true` or `false` state if we have to be definite. There is some part of the onboarding where this distinction matters though I can't remember where right now :) I'll go back and review it tomorrow to make sure it's clearer.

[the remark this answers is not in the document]
simonheys: On investigation, the onboarding ended up not using this hook after all. Have reverted the change

dhruvkelawala (approved):

PR #1237, opened 2022-09-27
feat: transaction review value

simonheys:
Show token and value of ERC20 approve and transfers in transaction review

image

### DIFF
--- a/packages/extension/src/shared/call/erc20ApproveCall.ts
+++ b/packages/extension/src/shared/call/erc20ApproveCall.ts
@@ -1,0 +1,41 @@
+import { Call } from "starknet"
+
+import { Erc20Call, parseErc20Call, validateERC20Call } from "./erc20Call"
+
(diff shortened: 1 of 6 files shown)


packages/extension/src/shared/call/erc20ApproveCall.ts
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Checks if the call has the expected shape of an ERC20 approve call
+ * @param call Call
+ * @returns true if the call passes validation
+ */
+
+export const isErc20ApproveCall = (call: Call): call is Erc20ApproveCall => {
+  try {
+    if (
+      call &&
```
dhruvkelawala, line 18:
```suggestion
```

Can be removed as `call` is a required parameter

dhruvkelawala (changes requested):

dhruvkelawala (approved):

PR #1281, opened 2022-10-05
fix: copy address

simonheys:
Adds a button which copies the checksummed address on click. Intercepts mouseup on the address field to select the whole text. Intercepts copy event to replace spaced address with checksummed.

https://user-images.githubusercontent.com/175607/194035093-09267903-7134-4962-83d1-9c6218d070f6.mov

### DIFF
--- a/packages/extension/src/ui/services/addresses.ts
+++ b/packages/extension/src/ui/services/addresses.ts
@@ -1,4 +1,3 @@
-import { isHexString } from "ethers/lib/utils"
 import {
   constants,
   getChecksumAddress,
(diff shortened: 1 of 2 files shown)


packages/extension/src/ui/services/addresses.ts
```diff
@@ -1,4 +1,3 @@
-import { isHexString } from "ethers/lib/utils"
```
simonheys, line 1: Unused import

dhruvkelawala (approved):

PR #1323, opened 2022-10-18
fix: use noble hash in ethers v5

janek26:


packages/extension/src/background/wallet.ts
```diff
@@ -280,9 +280,14 @@ export class Wallet {
       return true
     }
 
+    const throtteledProgressCallback = throttle(progressCallback ?? noop, 50, {
```
simonheys, line 283:
```suggestion
    const throttledProgressCallback = throttle(progressCallback ?? noop, 50, {
```

packages/extension/src/background/wallet.ts
```diff
@@ -280,9 +280,14 @@ export class Wallet {
...
+      leading: true,
+      trailing: true,
+    })
+
     // wallet is not initialized: let's initialise it
     if (!(await this.isInitialized())) {
-      await this.generateNewLocalSecret(password, progressCallback)
+      await this.generateNewLocalSecret(password, throtteledProgressCallback)
```
simonheys, line 290:
```suggestion
      await this.generateNewLocalSecret(password, throttledProgressCallback)
```

packages/extension/src/background/wallet.ts
```diff
@@ -296,7 +301,7 @@ export class Wallet {
       const wallet = await ethers.Wallet.fromEncryptedJson(
         backup,
         password,
-        progressCallback,
+        throtteledProgressCallback,
```
simonheys, line 304:
```suggestion
        throttledProgressCallback,
```

simonheys (approved):
This is an epic improvement. Small typo otherwise good to go.

EDIT: Looks like tests are failing in CI, will also check locally

PR #1324, opened 2022-10-18
feat: init design system

simonheys:
This sets up a sparse initial shared UI package in `packages/ui`, and moves the existing extension theme there.

Alongside the theme itself, the package includes some very simple shorthand utilities 'inspired' by the likes of chakra-ui. These are not anywhere near as expansive or flexible as chakra but allow shorthand access to theme values which should be all that is needed in 90% of cases, and should encourage use of theme values instead of e.g. locally coded pixel or font values.

Button and Typography components with variants have been created from the new design system designs. The variant names are the same as those in the design files. This means when building new screens you can see the variant names within the design and use those to access theme values or components. 

The goal eventually is to move all shared components into this package and gradually remove them from extension.

See the README for examples: https://github.com/argentlabs/argent-x/tree/feat/init-design-system/packages/ui


simonheys:
Closed after discussion in favour of #1325

PR #1325, opened 2022-10-19
feat: base design system package

simonheys:
Sets up a base design system package `@argent/ui` with theme and example components and stories for chakra-ui. This is a starting point for further development - the theme is not yet used in the extension.

See the package README for an example: https://github.com/argentlabs/argent-x/tree/feat/init-design-system-chakra/packages/ui


packages/extension/package.json
```diff
@@ -73,6 +73,7 @@
   },
   "dependencies": {
     "@argent/x-sessions": "^4.8.5",
+    "@argent-x/ui": "*",
```
janek26, line 76:
```suggestion
    "@argent/ui": "*",
```

Lets make it one argent

packages/extension/src/ui/App.tsx
```diff
@@ -1,3 +1,9 @@
+import {
+  FixedGlobalStyle,
+  ThemeProvider,
+  ThemedGlobalStyle,
+  muiTheme,
+} from "@argent-x/ui/src/theme"
```
janek26, line 6: I dont think that we should import from src in any case

There should be a build pipeline which takes care of bundling to `@argent/ui` or creates entrypoints like `@argent/ui/theme` which is also possible inside package.json using the exports syntax

packages/ui/package.json
```diff
@@ -0,0 +1,58 @@
...
+  },
+  "scripts": {
+    "gen:theme-typings": "chakra-cli tokens ./src/theme/chakraTheme.ts",
+    "lint": "eslint . --cache --ext .ts,.tsx",
+    "postinstall": "yarn gen:theme-typings"
+  },
+  "dependencies": {
+    "@chakra-ui/react": "^2.3.6",
+    "@emotion/react": "^11.10.4",
+    "@emotion/styled": "^11.10.4",
```
janek26, line 34: are both of the emotion libs needed?

packages/ui/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "@google/model-viewer": "^2.0.0",
+    "@mui/icons-material": "^5.3.1",
+    "@mui/material": "^5.1.0",
+    "@mui/styled-engine-sc": "^5.10.3",
```
janek26, line 38: can we drop mui in `@argent/ui`

packages/ui/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "@tippyjs/react": "^4.2.6",
+    "colord": "^2.9.2",
+    "framer-motion": "^7.6.1",
+    "lodash-es": "^4.17.21",
+    "qr-code-styling": "^1.6.0-rc.1",
+    "react": "^18.0.0",
```
janek26, line 44: should be peer dep

packages/ui/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "react-copy-to-clipboard": "^5.0.4",
+    "react-dom": "^18.0.0",
```
janek26, line 46: peer dep

packages/ui/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "react-dropzone": "^14.0.0",
+    "react-hook-form": "^7.33.0",
+    "react-measure": "^2.5.2",
+    "react-router-dom": "^6.0.1",
+    "react-select": "^5.4.0",
+    "react-textarea-autosize": "^8.3.4",
+    "styled-components": "^5.3.5",
+    "styled-normalize": "^8.0.7",
```
janek26, line 54: styled-* needed?

packages/ui/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "tracking-link": "^4.0.1",
```
janek26, line 55: we have our own impl now

packages/ui/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "qr-code-styling": "^1.6.0-rc.1",
+    "react": "^18.0.0",
+    "react-copy-to-clipboard": "^5.0.4",
+    "react-dom": "^18.0.0",
+    "react-dropzone": "^14.0.0",
+    "react-hook-form": "^7.33.0",
+    "react-measure": "^2.5.2",
+    "react-router-dom": "^6.0.1",
+    "react-select": "^5.4.0",
+    "react-textarea-autosize": "^8.3.4",
```
janek26, line 52: if I import a button from this package, I dont want to use all of these dependencies I dont need. I think we should put heavy focus on treeshaking here and use the package.json exports syntax to get different entrypoints

packages/ui/src/theme/GlobalStyle.tsx
```diff
@@ -0,0 +1,49 @@
+import { createGlobalStyle } from "styled-components"
```
janek26, line 1: I dont think this should be part of a ui library

packages/ui/src/theme/ThemeProvider.tsx
```diff
@@ -0,0 +1,12 @@
+import { FC, PropsWithChildren } from "react"
```
janek26, line 1: lets get rid of this file

packages/ui/src/theme/muiTheme.ts
```diff
@@ -0,0 +1,13 @@
+import { createTheme } from "@mui/material/styles"
```
janek26, line 1: lets remove this too

janek26 (approved):

packages/ui/package.json
simonheys: As discussed - will come back to this laters

janek26 (approved):
thanks a lot for this! 
I'll add turborepo to the mix with ww

PR #1255, opened 2022-09-30
Feat/starknet v0.10.0

dhruvkelawala:

PR #1328, opened 2022-10-19
Feat/starknet v0.10.1

dhruvkelawala:


dhruvkelawala:
To be merged after #1255

packages/extension/src/ui/features/actions/ApproveSignatureScreen.tsx
```diff
@@ -24,9 +28,24 @@ export const Pre = styled.pre`
...
+  const { feeTokenBalance } = useFeeTokenBalance(selectedAccount)
+
+  const shouldBeUpgraded = Boolean(needsUpgrade && feeTokenBalance?.gt(0))
+
+  if (selectedAccount?.needsDeploy && !selectedAccount.deployTransaction) {
+    return <DeployAccountScreen {...props} />
+  }
+
+  if (shouldBeUpgraded) {
+    return <UpgradeScreenV4 upgradeType="account" {...props} />
```
janek26, line 46: can you please use the `getNonce` check to determine if an upgrade is mandatory here too?
As there might be new upgrades in the future which are not mandatory and where we dont want to show this screen

dhruvkelawala: For single account, we can use that but what about the multicall that we are using here https://github.com/argentlabs/argent-x/blob/dc4aa73cc901626dde69f8b525439d5dc9555d60/packages/extension/src/ui/features/accounts/upgrade.service.ts#L61 

The multicall will throw the entire api call even if a single account doesn't have `get_nonce` selector. So we can ideally use it to `checkIfV4UpgradeAvailableOnNetwork` but we can't determine if the account has fees for upgrade which is also needed for truthy value.

simonheys (approved):

PR #1342, opened 2022-10-21
feat: icons from Figma

simonheys:
Adds a script `yarn gen:icons` which fetches icon SVGs from a Figma frame and converts them into react components. These components take the current font size and color.

image


packages/ui/scripts/generate-icons.ts
```diff
@@ -0,0 +1,163 @@
...
+import fetch from "cross-fetch"
+import * as dotenv from "dotenv"
+import { camelCase, upperFirst } from "lodash"
+
+dotenv.config()
+
+const OUTPUT_FOLDER = path.join(__dirname, "../src/components/icons")
+
+const FIGMA_ACCESS_TOKEN = process.env.FIGMA_ACCESS_TOKEN
+const FIGMA_ICONS_FILE_KEY = "LHwepHSS4bouYQjbMOZJjW"
```
simonheys, line 14: This is useless without the access token in dotenv. But could potentially also move into dotenv file as well

janek26 (approved):
genius! I hope we can use that in the future, even if the icons source of truth moves to a different figma file with a different structure

PR #1374, opened 2022-11-01
fix: don't show empty screen when closing account list

simonheys:
Fixes an issue where the new account list has a close button that may cause user to see an empty screen. This state will not be possible once the full UI implementation work is completed.


janek26 (approved):
I think whatever gets merged into develop should be standalone and ready to get released to the user, no?
If that's different from what you expected from develop, we should probably come up with a definition of done or sth

PR #1375, opened 2022-11-02
feat: update network switcher

simonheys:
Updates the network switcher UI, adds new 'No account' screen and decouples switching network from account.

In this PR, switching network will try and switch to the first account on that network, or show the empty screen if there isn't one. A future enhancement will remember the previously selected account on each network and select that automatically.

The network switcher uses a shared Chakra Menu component which is themed in the `@argent/ui` package.

https://user-images.githubusercontent.com/175607/199512708-06b91a6c-960b-4282-b74e-05dd61b9f41e.mov


simonheys:
Moved to draft while investigating an issue with account names

packages/extension/src/ui/features/accounts/HideOrDeleteAccountConfirmScreen.tsx
```diff
@@ -5,13 +5,14 @@ import styled from "styled-components"
...
 import { formatFullAddress } from "../../services/addresses"
 import { deleteAccount } from "../../services/backgroundAccounts"
 import { P } from "../../theme/Typography"
 import { ConfirmScreen } from "../actions/ConfirmScreen"
-import { recover } from "../recovery/recovery.service"
 import { getAccountName, useAccountMetadata } from "./accountMetadata.state"
 import { useAccount } from "./accounts.state"
+import { autoSelectAccountOnNetwork } from "./switchAccount"
 
 const StyledP = styled(P)`
```
bluecco, line 17: this could use the new ui library instead of styled components

packages/extension/src/ui/features/networks/NetworkSwitcher.tsx
```diff
@@ -1,96 +1,27 @@
...
-        border-radius: 15px 15px 0 0;
-      }
-
-      &:hover ${NetworkName} {
-        min-width: 110px;
-      }
-    `}
-`
-
 export const NetworkStatusWrapper = styled.div`
```
bluecco, line 25: can be updated with chakra component (removing also styled-components import)

packages/extension/src/ui/features/accounts/HideOrDeleteAccountConfirmScreen.tsx
simonheys: Good spot - this screen will be changed in a future PR

packages/extension/src/ui/features/accounts/AccountListItem.tsx
```diff
@@ -38,6 +37,13 @@ type AccountListItemWrapperProps = Pick<
   dark?: boolean
 }
 
+export const NetworkStatusWrapper = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: right;
+  gap: 4px;
+`
```
janek26, line 45: why no chakra `Flex` component? :p :)

janek26 (approved):
well done! Does it still show two `+` icons (top right and bottom of list)?
If so, should we remove the one from the bottom of the list?

packages/extension/src/ui/features/accounts/AccountListItem.tsx
simonheys: Nice catch! These will be replaced in the next PR!

PR #1377, opened 2022-11-02
BLO-530: Implement Lockscreen redesign

bluecco:


packages/extension/src/ui/features/lock/LockScreen.tsx
```diff
@@ -1,97 +1,98 @@
...
+            <>
+              <StickyGroup>
+                <Button
+                  gap="2"
+                  colorScheme="primary"
+                  type="submit"
+                  disabled={!isDirty || isSubmitting}
+                  width="100%"
+                >
+                  {isLoading && (
```
janek26, line 83: chakra button component ships with buildin `isLoading` prop! :) 
https://chakra-ui.com/docs/components/button/props

bluecco: nice, will update the component

bluecco: updated @janek26

janek26 (approved):
works like a charm!

PR #1398, opened 2022-11-04
chore: reduce network requests related to accounts

bluecco:
10 secs as placeholder, let's define a better timeout


janek26:
how does that reduce network requests? Doesnt it just add one network request every x seconds?

bluecco:
@janek26  it is true that add a call after the interval elapsed
but also, during that interval, the call is not performed (that's what I've observed with manual testing)

with swr I couldn't find other options to optimize it

otherwise a refactor on the component could be done, but need a little of design before implementing

bluecco:
timeout updated @janek26

janek26 (approved):

PR #1401, opened 2022-11-07
feat: light mode

simonheys:


packages/ui/src/components/Button.tsx
```diff
@@ -56,26 +62,52 @@ export const buttonTheme = defineStyleConfig({
...
+            bg: mode("gray.100", "gray.900")(props),
           },
           _active: {
-            bg: "gray.800",
+            bg: mode("gray.200", "gray.800")(props),
           },
         }
+      } else if (c === "neutrals") {
+        return {
+          bg: mode(`white`, `${c}.800`)(props),
```
janek26, line 80: instead of using mode, shouldnt we use the "pseudo classes" `_dark` and `_light`? (Less JS and more CSS, for SSR)

janek26: https://chakra-ui.com/docs/styled-system/semantic-tokens#conditional-tokens

PR #1403, opened 2022-11-08
feat: remember last selected account on network

bluecco:
when switching network, keeps the current account if it exist on the network
otherwise, open the first account

### DIFF
--- a/packages/extension/e2e/switchNetwork.test.ts
+++ b/packages/extension/e2e/switchNetwork.test.ts
@@ -1,0 +1,40 @@
+import { expect } from "@playwright/test"
+
+import { test } from "./fixture"
+import { getAccountAddressFromAccountPage } from "./selectors/getAccountAddressFromAccountPage"
+import { getBalanceFromAccountPage } from "./selectors/getBalanceFromAccountPage"
+import { setupNewAccountWithTestnetEth } from "./setups/accountWithTestnetEth"
+import { navigateFromAccountToAccountList } from "./steps/navigateFromAccountToAccountList"
+import { newAccount } from "./steps/newAccount"
+import { selectAccountFromAccountList } from "./steps/selectAccountFromAccountList"
+import { switchNetwork } from "./steps/switchNetwork"
+import { dismissUserReview } from "./steps/userReview"
+
+test("switch network and keep account", async ({ page, context }) => {
+  const { address: address1_localhost } = await setupNewAccountWithTestnetEth(
+    page,
+    context,
+  )
+
+  await navigateFromAccountToAccountList(page)
+  await newAccount(page)
+  const address2_localhost = await getAccountAddressFromAccountPage(page)
+  const b2 = await getBalanceFromAccountPage(page, "Ethereum")
+  expect(b2).toBe("0.0")
+  await navigateFromAccountToAccountList(page)
+  await selectAccountFromAccountList(page, address1_localhost)
+  await dismissUserReview(page)
+
+  await switchNetwork(page, "Testnet")
+  const address1_goerli = await getAccountAddressFromAccountPage(page)
+
+  await switchNetwork(page, "Localhost")
+  expect(address1_localhost).toEqual(address1_goerli)
+
+  await navigateFromAccountToAccountList(page)
+  await selectAccountFromAccountList(page, address2_localhost)
+
+  await switchNetwork(page, "Testnet")
+  const currentGoerli = await getAccountAddressFromAccountPage(page)
+  expect(currentGoerli).toEqual(address1_goerli)
+})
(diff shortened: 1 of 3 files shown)


packages/extension/src/ui/features/accounts/switchAccount.ts
```diff
@@ -24,10 +26,18 @@ export const autoSelectAccountOnNetwork = async (networkId: string) => {
...
-    return selectedAccount
+    const existingAccoutnOnNetwork = selectedAccount
+      ? visibleAccountsOnNetwork.find((account) =>
+          addressesEquals(account, selectedAccount),
+        )
+      : null
+
+    // if the selected account is not on the network, switch to the first visible account
+    useSelectedAccountStore.setState({
+      selectedAccount: existingAccoutnOnNetwork || visibleAccountsOnNetwork[0],
```
simonheys, line 39:
```suggestion
      selectedAccount: existingAccountOnNetwork || visibleAccountsOnNetwork[0],
```

packages/extension/src/shared/wallet.service.ts
```diff
@@ -17,13 +17,20 @@ export const isDeprecated = ({ signer, network }: WalletAccount): boolean => {
...
   try {
-    return (
-      BigNumber.from(a.address.toLowerCase()).eq(
-        BigNumber.from(b.address.toLowerCase()),
-      ) && a.networkId === b.networkId
+    return BigNumber.from(a.address.toLowerCase()).eq(
+      BigNumber.from(b.address.toLowerCase()),
     )
+  } catch (e) {
+    console.error("~ addressesEquals", e)
```
simonheys, line 26: Is this console.error needed?

packages/extension/src/shared/wallet.service.ts
```diff
@@ -17,13 +17,20 @@ export const isDeprecated = ({ signer, network }: WalletAccount): boolean => {
   )
 }
 
-export const accountsEqual = (a: BaseWalletAccount, b: BaseWalletAccount) => {
+export const addressesEquals = (a: BaseWalletAccount, b: BaseWalletAccount) => {
```
simonheys, line 20: Since we already have a function elsewhere called `isEqualAddress`, I think we should differentiate this better and maybe call it `isEqualWalletAddress` or something like that, just so the use case is clear it's about wallets

packages/extension/src/shared/wallet.service.ts
```diff
@@ -17,13 +17,20 @@ export const isDeprecated = ({ signer, network }: WalletAccount): boolean => {
...
   try {
-    return (
-      BigNumber.from(a.address.toLowerCase()).eq(
-        BigNumber.from(b.address.toLowerCase()),
-      ) && a.networkId === b.networkId
+    return BigNumber.from(a.address.toLowerCase()).eq(
+      BigNumber.from(b.address.toLowerCase()),
```
simonheys, line 23: Could this use the existing `isEqualAddress` function, or is different and special somehow?

simonheys:
Small thing, can change the PR name to conventional commits style so something like 'feat: remember last selected account on network'

packages/extension/e2e/switchNetwork.test.ts
```diff
@@ -0,0 +1,40 @@
...
+import { getBalanceFromAccountPage } from "./selectors/getBalanceFromAccountPage"
+import { setupNewAccountWithTestnetEth } from "./setups/accountWithTestnetEth"
+import { navigateFromAccountToAccountList } from "./steps/navigateFromAccountToAccountList"
+import { newAccount } from "./steps/newAccount"
+import { selectAccountFromAccountList } from "./steps/selectAccountFromAccountList"
+import { switchNetwork } from "./steps/switchNetwork"
+import { dismissUserReview } from "./steps/userReview"
+
+test("switch network and keep account", async ({ page, context }) => {
+  const { address: address1_localhost } = await setupNewAccountWithTestnetEth(
```
simonheys, line 14: I don't think the code uses `snake_case` anywhere else (unless it's for contract methods or API keys which are outside our control) so `address1_localhost` -> `address1Localhost` or `localhostAddress1` would be more consistent here

simonheys (changes requested):
Works for me, added a few comments about style and typos

packages/extension/src/shared/wallet.service.ts
bluecco: I wasn't aware of `isEqualAddress` :)
so I divided just moved this part out from `accountsEqual`

packages/extension/src/shared/wallet.service.ts
bluecco: I just kept the error when I divided the function
but I can remove it, what do you think?

bluecco:
@simonheys pr updated

simonheys (approved):

PR #1420, opened 2022-11-11
feat: smart contract declare and deploy - skeleton ui

bluecco:
settings page new UI and skeleton for developer settings

will point and merged on `develop` after `feat/starknet-v0.10.1` will be released


packages/extension/src/ui/AppRoutes.tsx
```diff
@@ -184,6 +188,22 @@ const legacyUiWalletRoutes = (
     <Route path={routes.addPlugin.path} element={<AddPluginScreen />} />
     <Route path={routes.settings.path} element={<SettingsScreen />} />
     <Route path={routes.settingsSeed.path} element={<SeedSettingsScreen />} />
+    <Route
```
simonheys, line 191: If these new routes are using the new UI stuffs, they can be moved out of the `legacyUiWalletRoutes` and into `walletRoutes` - it has a much simpler, responsive wrapper. Would be interested to know if you see any issues when doing this too.

bluecco: ah that's good to know!

I will implement that in the next pr (where I'm adding the select)

bluecco:
image

image

image

packages/extension/src/ui/features/settings/DeveloperSettings.tsx
```diff
@@ -0,0 +1,44 @@
...
+const DeveloperSettings: FC = () => {
+  const navigate = useNavigate()
+  return (
+    <>
+      <NavigationBar
+        leftButton={<BarBackButton onClick={() => navigate(-1)} />}
+        title={"Developer Settings"}
+      />
+
+      <Stack direction="column" mx="4">
```
simonheys, line 19: Should probably use the shared `CellStack`, `ButtonCell` components for this screen (it is v v similar to https://github.com/argentlabs/argent-x/blob/02f6525ccb123ab26a349e56c9145bc1daa32ba8/packages/extension/src/ui/features/accountEdit/AccountEditScreen.tsx#L115) . I can see maybe needs a `LinkCell` may be as simple as `LinkCell = ... <ButtonCell as={Link} ... />` where `Link` is from react router (dom?)

bluecco: as said on slack, `Link` need the router context
for now, it won't be part of the `ui`

packages/extension/src/ui/features/settings/SettingsMenuItem.tsx
```diff
@@ -0,0 +1,34 @@
...
+}) => (
+  <Link to={to} onClick={onClick}>
+    <ButtonCell
+      leftIcon={<>{leftIcon}</>}
+      rightIcon={<>{rightIcon}</>}
+      width="100%"
+    >
+      {title}
+    </ButtonCell>
+  </Link>
```
simonheys, line 31: Should be able to just do this to get the ButtonCell to render as a Link;

```
<ButtonCell
    as={Link}
   ...combined props
  >
    {title}
  </ButtonCell>

File: packages/extension/src/ui/features/settings/SettingsMenuItem.tsx [hunk 4]
```diff
@@ -0,0 +1,34 @@
+import { ButtonCell, H6, icons } from "@argent/ui"
+import { Flex, Text } from "@chakra-ui/react"
```
simonheys, line 2: Some unused imports here?

bluecco: yep, forgot to remove them

packages/extension/src/ui/features/settings/SettingsScreen.tsx
```diff
@@ -61,186 +80,81 @@ export const SettingsScreenWrapper = styled.div`
...
-        </SettingsLinkItem>
-        <hr />
-        {!extensionIsInTab && (
-          <>
-            <SettingsLinkItem
+      <NavigationContainer
+        leftButton={<BarBackButton onClick={() => navigate(-1)} />}
+        title={"Settings"}
+      >
+        <Box mx="4" my="4">
```
simonheys, line 96: Can move the `AccountListScreenItem` inside the CellStack so this doesn't leave additional empty spacing when its not present? There is also the dedicated `SpacerCell` so we don't start defining local spacers everywhere; imagining something like

```
{account && <><AccountListScreenItem account={account} /><SpacerCell /></>}
```

packages/extension/src/ui/features/settings/SettingsScreen.tsx
```diff
@@ -61,186 +81,85 @@ export const SettingsScreenWrapper = styled.div`
...
+          {isPrivacySettingsEnabled && (
+            <SettingsMenuItem
+              leftIcon={<ShieldIcon />}
+              to={routes.settingsPrivacy()}
+              title="Privacy"
+            />
+          )}
+          <Link onClick={stopSession} to={routes.lockScreen()}>
+            <SpacerCell />
+            <Flex
```
simonheys, line 147: Maybe for the next PR but Chakra has a `Center` component could be used here

simonheys (approved):

PR #1433, opened 2022-11-14
Feat/support-new-multicall

dhruvkelawala:


dhruvkelawala:
Should I make tweaks to maxBatchSize and batchInterval or should we use the default ones for fetching TokenBalances and Account Implementations? @janek26

dhruvkelawala:
The current values looks ok to me

janek26 (approved):
good initial PR!
I think we should follow up with a PR sharing the upgrade state between account view and account list view
In my mind that's done by having single call hooks/functions only (getUpgradeStatus instead of getUpgradeStatuse**s**) so that we can use SWR to share the cache of the hook between single account view and accounts list view

PR #1434, opened 2022-11-15
fix/wallet_switchStarknetChain RPC Message

dhruvkelawala:


dhruvkelawala:
TODO
- [ ] Fix NetworkSwitcher
- [ ] Remove Console Logs
- [ ] Test thoroughly

packages/extension/src/background/actionHandlers.ts
```diff
@@ -112,16 +115,62 @@ export const handleActionApproval = async (
...
+            `No accounts found on network with chainId ${action.payload.chainId}`,
+          )
+        }
+
+        return {
+          type: "APPROVE_REQUEST_SWITCH_CUSTOM_NETWORK",
+          data: { actionHash, selectedAccount },
+        }
+      } catch (error) {
+        console.error(error)
```
bluecco, line 169: can be removed?

packages/extension/src/background/actionHandlers.ts
```diff
@@ -112,16 +115,61 @@ export const handleActionApproval = async (
...
+          (n) => n.chainId === action.payload.chainId,
+        )
+
+        if (!network) {
+          throw Error(
+            `Network with chainId ${action.payload.chainId} not found`,
+          )
+        }
+
+        const accountsOnNetwork = await getAccounts(
```
bluecco, line 146: @dhruvkelawala was wondering, does it makes sense to reject if there are no accounts?
couldn't the user create it after switching to the custom network?

dhruvkelawala: This RPC message will be sent by dApp to ArgentX and not the user. For example, if a dApp wants to use TESTNET2 instead of TESTNET, it will send this message through window object. And when switching the network, the dApp will require an account to connect with. So if the account is not present, dApp cannot connect to anything and even ArgentX won't be able to maintain a state where no account is selected on a network. If a user wants to create an account on network, they can still do it directly from extension. This PR doesn't change that.

bluecco (approved):

dhruvkelawala:
@janek26 Waiting for your review on this

janek26 (approved):

PR #1436, opened 2022-11-15
feat: token list item UI

simonheys:


packages/extension/src/ui/features/accountTokens/TokenIconDeprecated.tsx
```diff
@@ -0,0 +1,40 @@
+import { FC } from "react"
+import styled from "styled-components"
+
+import { getTokenIconUrl } from "./TokenIcon"
+
+const Icon = styled.img<{ size: number }>`
```
bluecco, line 6: keeping this because this component will be removed going forward?

packages/extension/src/ui/features/accountTokens/TokenListItemDeprecated.tsx
```diff
@@ -0,0 +1,68 @@
+import styled, { css, keyframes } from "styled-components"
+
+export const TokenWrapper = styled.div`
```
bluecco, line 3: same as above, keeping `styled` since this component will be removed?

packages/extension/src/ui/features/accountTokens/TokenIconDeprecated.tsx
simonheys: Exactly, it will be removed naturally at some point. Beyond the scope of this PR

packages/extension/src/ui/features/accountTokens/TokenListItemDeprecated.tsx
simonheys: Exactly, it will be removed naturally at some point. Beyond the scope of this PR - this will probably be gone once we have updated the activity screen I think.

bluecco (approved):
looking good!

PR #1439, opened 2022-11-16
chore: memoize creation of provider for each baseUrl

simonheys:
This PR makes an adjustment so that calls to `getProvider()` are memoized and create a single instance of `SequencerProvider()` for each `baseUrl`, rather than creating a new instance of `SequencerProvider()` each time it is called.


dhruvkelawala (approved):
Nice one!

PR #1447, opened 2022-11-17
Feat: use multicall for individual token balances

simonheys:
This PR implements fetching individual token balances using Multicall package and adds a simple additional script `yarn devnet:setup-contracts` which will declare and deploy the Multicall contract to devnet using the UDC. Localhsot network config and e2e tests are updated so they are now using Multicall.


scripts/devnet-setup-contracts.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  if (!multicallContractClassHash) {
+    throw "Cannot continue without Multicall contract class hash"
+  }
+
+  console.log("Deploying Multicall contract...")
+  await universalDeployerDeployContract(multicallContractClassHash)
+
+  console.log(
+    "Multicall contract deployed… address should be 0x05754af3760f3356da99aea5c3ec39ccac7783d925a19666ebbeca58ff0087f4",
```
simonheys, line 20: The address is not returned, though apparently it is possible to get it from on-chain events.

packages/extension/src/ui/features/accountTokens/TokenListItemMulticall.tsx
```diff
@@ -0,0 +1,86 @@
...
+  token,
+  account,
+  ...rest
+}) => {
+  const { balanceOrError, isValidating } = useTokenBalanceForAccount(
+    token,
+    account,
+    {
+      suspense: true,
+      ...withPolling(30 * 1000) /** 30 seconds */,
```
janek26, line 39: can we do min 1 minute? A network request can already take up to 30s

packages/extension/src/ui/features/accountTokens/TokenListItemMulticall.tsx
```diff
@@ -0,0 +1,86 @@
...
+    },
+  )
+  const { tokenWithBalance, errorMessage } = useMemo(() => {
```
janek26, line 42: shoudnt this logic be inside `useTokenBalanceForAccount ` ? as it's very likely that you want to keep this logic when using useTokenBalanceForAccount somewhere else, no?

packages/extension/src/shared/multicall/getTokenBalance.ts
```diff
@@ -0,0 +1,24 @@
...
+import { getMulticallForNetwork } from "./getMulticall"
+
+/** use Multicall to get balance of token at account address on network */
+
+export const getTokenBalanceForWalletAccount = async (
+  token: BaseToken,
+  account: BaseWalletAccount,
+) => {
+  const network = await getNetwork(account.networkId)
+  const multicall = getMulticallForNetwork(network)
```
janek26, line 15: what happens if no multicall contract is deployed in the network?

I think we should detect if the multicall is deployed, and if not we should deploy it from within the extension instead of with a dedicated cli command, wdyt?

packages/extension/src/ui/features/accountTokens/TokenListItemContainer.tsx
```diff
@@ -0,0 +1,56 @@
...
+      {
+        token,
+        account,
+        shouldReturnError:
+          true /** using Suspense, causes error to be returned as `balance` instead of throwing */,
+      },
+      {
+        suspense:
+          true /** Suspense allows us to show an initial loader for all tokens */,
+        ...withPolling(60 * 1000) /** 30 seconds */,
```
simonheys, line 36:
```suggestion
        ...withPolling(60 * 1000) /** 60 seconds */,
```

janek26 (approved):
a masterpiece
we have 1 network request per screen now

we could share cache (ie balances in send screen for max button) to get to even less network requests

PR #1453, opened 2022-11-18
Feat/update wallet recovery

dhruvkelawala:


janek26:
hey, as agreed in PM lets close this task for now. Discovery is super hard due to so many combinations, I think we'll need something in the backend soon...

Have a nice weekend and go outside <3

PR #1466, opened 2022-11-22
feat: update activity cells

simonheys:
This PR updates the UI inside the cells of the activity view, and removes the old icons and components.

Screenshot 2022-11-22 at 16 52 56

Screenshot 2022-11-22 at 16 53 20


packages/ui/src/theme/index.tsx
```diff
@@ -42,6 +43,7 @@ const extendedTheme = extendTheme({
   },
   breakpoints,
   ...typography,
+  sizes,
```
simonheys, line 46: Added so the theme can add additional spaces, otherwise the `sizes` file contains the existing default values from chakra

PR #1489, opened 2022-11-25
BLO 584 - fix: settings page back button navigate to account tokens page

bluecco:


packages/extension/src/ui/features/settings/SettingsScreen.tsx
```diff
@@ -92,7 +92,9 @@ export const SettingsScreen: FC = () => {
   return (
     <>
       <NavigationContainer
-        leftButton={<BarBackButton onClick={() => navigate(-1)} />}
+        leftButton={
+          <BarBackButton onClick={() => navigate(routes.accountTokens())} />
```
janek26, line 96: you can enter settings from the account list aswell, right?
In that case you would not go back to account list, but go inside the account

we can spin this out to another ticket if you want, this is better than having a dead end

janek26 (approved):

packages/extension/src/ui/features/settings/SettingsScreen.tsx
bluecco: not from the account list, it goes here
image

just checked now and it should be safe

PR #1517, opened 2022-12-01
BLO-587: use safeTransferFrom when sending erc721 NFTs

bluecco:


packages/extension/src/shared/call/erc721TransferCall.ts
```diff
@@ -11,16 +11,16 @@ export interface Erc721TransferCall extends Call {
...
 
-export const isErc721TransferCall = (
+export const isErc721SafeTransferCall = (
   call: Call,
 ): call is Erc721TransferCall => {
   try {
     if (
       call.contractAddress &&
-      call.entrypoint === "transferFrom" &&
+      call.entrypoint === "safeTransferFrom" &&
```
janek26, line 23: shouldnt we support both?
As I understand it this is the parser to "pretty show" the action, no? I think both could be shown the same way, if a dapp ie uses the old transfer

bluecco: mmm ticket says 
`Use safeTransferFrom to send NFTs instead of trasferFrom`

that's why I've replaced it, maybe we can discuss this @janek26

janek26: right,`Use safeTransferFrom to send NFTs instead of trasferFrom` would have been as easy as 
https://github.com/argentlabs/argent-x/pull/1517/files#diff-65514291c3def75d3087efce0068f2dfef666124bbc047aa5caf428ad4b94190R165-R172
no?

Obviously we want to show it pretty aswell, so I would've just added the `call.entrypoint === "safeTransferFrom` condition into erc721Transfer.ts,without replacing/renaming `isErc721TransferCall` at all

It makes sense to show both methods the same, no? maybe an additional indicator which method will be used

packages/extension/src/shared/call/erc721TransferCall.ts
```diff
@@ -11,7 +11,7 @@ export interface Erc721TransferCall extends Call {
     tokenIdLowFelt: number.BigNumberish,
     tokenIdHighFelt: number.BigNumberish,
   ]
-  entrypoint: "transferFrom"
+  entrypoint: "safeTransferFrom"
```
janek26, line 14:
```suggestion
  entrypoint: "safeTransferFrom" | "transferFrom"
```

janek26 (approved):

PR #1528, opened 2022-12-05
BLO-598 - update account list UI

simonheys:
Updates the upgrade and connected states of account list items

After
Screenshot 2022-12-05 at 15 23 26

Before
Screenshot 2022-12-05 at 14 19 06


janek26 (approved):
looks nice!

PR #1535, opened 2022-12-06
BLO-581 - fix(ui): improve handling of individual token errors

simonheys:
Use Promise.allSettled so a single failure doesn't break whole list

Screenshot 2022-12-06 at 11 41 48


janek26 (approved):
noice!

PR #1554, opened 2022-12-08
BLO-605 - feat(ui): opt-in scroll restoration

simonheys:
In preparation for animated transition work, this PR adds opt-in scroll restoration to `NavigationContainer` and a hook for using it independently. To simplify this initial implementation, scroll is restored only on 'pop' (back) navigation.

### DIFF
--- a/packages/ui/src/hooks/useScroll.ts
+++ b/packages/ui/src/hooks/useScroll.ts
@@ -1,8 +1,12 @@
 import { useCallback, useRef, useState } from "react"
 
-export interface IScroll {
+export interface ScrollProps {
   scrollTop: number
   scrollLeft: number
+}
+
+export interface UseScrollProps {
+  onScroll?: (scroll: ScrollProps) => void
 }
 
 /**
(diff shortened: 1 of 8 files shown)


packages/ui/src/hooks/useScroll.ts
```diff
@@ -51,5 +59,9 @@ export const useScroll = () => {
     [onScroll],
   )
```
janek26, line 61: can we replace [this](https://github.com/argentlabs/argent-x/pull/1554/files#diff-794108ed6d5e63960e42d53a058ceb2d4db5628ae43c4a1da38df8c352cf0661R56)
```
ref.current.addEventListener("scroll", onScroll)
```
with the passive version?

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#improving_scrolling_performance_with_passive_listeners

packages/ui/src/hooks/useScrollRestoration.ts
```diff
@@ -0,0 +1,47 @@
...
+      storage.set(storageKey, scroll)
+    },
+    [storage, storageKey],
+  )
+  const useScrollProps = useScroll({ onScroll })
+  const navigationType = useNavigationType()
+
+  useLayoutEffect(() => {
+    if (navigationType === NavigationType.Pop) {
+      const value = storage.get(storageKey)
```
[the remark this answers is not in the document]
simonheys: Yes, it is stored in memory by default https://github.com/argentlabs/argent-x/blob/7d7e7db68734979052875d09604ddde8b2a0066d/packages/ui/src/hooks/useScrollRestoration.ts#L13-L14

packages/ui/src/hooks/useScroll.ts
simonheys: Added the `passive`. Re debouncing, this hook is also used to update the 'navigation bar' background colour when the user scrolls, so should keep the default refresh rate.

janek26 (approved):

PR #1572, opened 2022-12-20
BLO-617 transitions

simonheys:
Adds package `stack-router` and uses it in place of `react-router-dom` in the extension. Original imports left in place to make this easy to revert.


packages/extension/src/ui/AppRoutes.tsx
```diff
@@ -256,66 +294,95 @@ const walletRoutes = (
 const fullscreenRoutes = (
   <>
     <Route
+      presentation="replace"
```
janek26, line 297: does it make sense to have this as default?

janek26 (approved):

packages/extension/src/ui/AppRoutes.tsx
simonheys: Yes definitely

PR #1587, opened 2023-01-03
BLO-636 - fix: handle transfer event with no contract address

simonheys:


packages/extension/src/ui/features/accounts/accounts.service.ts
```diff
@@ -46,7 +46,12 @@ export const getAccountImageUrl = (
 }
 
 export const stripAddressZeroPadding = (accountAddress: string) => {
-  return number.toHex(toBN(number.hexToDecimalString(accountAddress)))
+  try {
+    return number.toHex(toBN(number.hexToDecimalString(accountAddress)))
+  } catch {
+    // ignore parsing erros
```
simonheys, line 52:
```suggestion
    // ignore parsing errors
```

PR #1592, opened 2023-01-04
fix: avoid retrieve accounts that are already in store

bluecco:


packages/extension/src/background/wallet.ts
```diff
@@ -384,9 +389,11 @@ export class Wallet {
 
     const network = await this.getNetwork(networkId)
 
+    console.log(await this.walletStore.get(withHiddenSelector))
     await this.discoverAccountsForNetwork(network, 1) // discover until there is an free index found
 
     const accounts = await this.walletStore.get(withHiddenSelector)
+    console.log(accounts)
```
simonheys, line 396:
```suggestion
```

bluecco: oopsy

packages/extension/src/background/wallet.ts
```diff
@@ -366,14 +366,19 @@ export class Wallet {
...
+    const walletStoreAccounts = await this.walletStore.get()
+    const exists = accounts.some((a) =>
+      walletStoreAccounts.find((w) => isEqualAddress(w.address, a.address)),
+    )
 
-    await this.walletStore.push(accounts)
+    if (!exists) {
+      await this.walletStore.push(accounts)
+    }
   }
```
dhruvkelawala, line 382: This doesn't make sense to me somehow. Because this function will never update walletStore unless it's empty. And if that's the case, we can remove the function call from `newAccount`

bluecco: hey @dhruvkelawala just to be sure, you're suggesting to remove `await this.discoverAccountsForNetwork(network, 1)` right?

I've already tested without that line and works as exepected

dhruvkelawala: Yes, that's what I meant. Hopefully removing that doesn't introduce any new bugs. Can you run e2e tests locally to ensure this?

bluecco: already ran and everything good
image

dhruvkelawala (approved):

dhruvkelawala:
Okay then let's push that so that we can merge it

dhruvkelawala (approved):

PR #1613, opened 2023-01-10
Blo 568

bluecco:

Fix how fee estimation is calculated in case a deploy account is needed.

### Changes
- use `estimateFeeBulk` method to calculate deploy account + transaction estimate fees

### Checklist

- [x] Code self-reviewed
- [x] Code self-tested


packages/extension/src/background/transactions/transactionExecution.ts
```diff
@@ -107,11 +114,19 @@ export const executeTransactionAction = async (
...
-    )
-    maxFee = number.toHex(stark.estimatedFeeToMaxFee(suggestedMaxFee, 1))
+    if ("estimateFeeBulk" in starknetAccount) {
+      const bulkTransactions: TransactionBulk = [
+        {
+          type: "INVOKE_FUNCTION",
+          payload: transactions,
+        },
+      ]
+      const estimateFeeBulk = await starknetAccount.estimateFeeBulk(
```
dhruvkelawala, line 124: I would recommend that `estimateFeeBulk` is done before Account is deployed and maxFee with overhead should be sent when doing `wallet.deployAccount`. At this position, it doesn't make any sense. I haven't tested this yet.. which I will now, but I think it won't work as well because "The account is not deployed". `estimate_fee_bulk` uses Starknet State without committing, so it exist only for that call.

dhruvkelawala:
I was right :) 

Screenshot 2023-01-12 at 11 31 47

dhruvkelawala (changes requested):
.

packages/extension/src/background/utils/argentMaxFee.ts
```diff
@@ -0,0 +1,4 @@
+import { number, stark } from "starknet"
+
+export const argentMaxFee = (suggestedMaxFee: number.BigNumberish) =>
```
dhruvkelawala, line 3: Just a minor thing but can we add documentation for this function? It will be very useful

dhruvkelawala (approved):

PR #1617, opened 2023-01-11
BLO-583 feat: 2FA

simonheys:

Adds initial implementation of 2FA with feature flag, enable by setting `FEATURE_ARGENT_SHIELD=true` in your local `.env`

### Changes

- add 'guardian' flag to accounts
- add detection of accounts with guardian on chain
- add GuardianSigner used for accounts with guardian set
- add 2FA screens

### Checklist

- [x] Code self-reviewed
- [x] Code self-tested


packages/extension/src/background/GuardianSigner.ts
```diff
@@ -0,0 +1,143 @@
...
+import { cosignerSign } from "../shared/shield/backend/account"
+import { getVerifiedEmailIsExpired } from "../shared/shield/verifiedEmail"
+
+export interface CosignerMessage {
+  message: any
+  type: "starknet" | "starknetDeploy"
+}
+
+export class GuardianSigner extends Signer {
+  constructor(keyPair: KeyPair) {
```
janek26, line 25: to make this more reuseable, can we make it more stateless?

like pass the JWT/baseUrl inside the constructor in example, so we can use pure functions only in the implementation

packages/extension/src/background/GuardianSigner.ts
```diff
@@ -0,0 +1,143 @@
...
+    super(keyPair)
+  }
+
+  public async cosignMessage(
+    cosignerMessage: CosignerMessage,
+  ): Promise<Signature> {
```
janek26, line 31: should probably just reuse the signature for `signTransaction`

packages/extension/src/background/accountMessaging.ts
```diff
@@ -177,6 +179,34 @@ export const handleAccountMessage: HandleMessage<AccountMessage> = async ({
...
+        await actionQueue.push({
+          type: "TRANSACTION",
+          payload: {
+            transactions: {
+              contractAddress: account.address,
+              entrypoint: "changeGuardian",
+              calldata: [number.hexToDecimalString(guardian || constants.ZERO)],
+            },
+            meta: {
+              isChangeGuardian: true,
```
janek26, line 195: we're getting more and more of these flag fields. Maybe we should unite them under one union type called `type`?

packages/extension/src/background/miscellaneousMessaging.ts
```diff
@@ -20,6 +22,9 @@ export const handleMiscellaneousMessage: HandleMessage<
         browser.storage.sync.clear()
         browser.storage.managed.clear()
         browser.storage.session.clear()
+        if (ARGENT_SHIELD_ENABLED) {
+          await resetDevice()
+        }
```
janek26, line 27: shouldnt we simply do that always?

packages/extension/src/background/wallet.ts
```diff
@@ -643,17 +655,27 @@ export class Wallet {
         : await this.getNetwork(selector.networkId),
     )
 
+    const useGuardianSigner = ARGENT_SHIELD_ENABLED && account.guardian
```
janek26, line 658: `use` makes me always think `hooks` :D

packages/extension/src/background/wallet.ts
```diff
@@ -643,17 +655,27 @@ export class Wallet {
...
+
+    const keyPairOrSigner = useGuardianSigner
+      ? new GuardianSigner(keyPair)
+      : keyPair
```
janek26, line 662: should we have a more general `getSigner` method?
instead if providing a keyPair we can also provide a default signer with that keypair, so I'm thinking `function getSignerForAccount(account: BaseAccount): SignerInterface`

packages/extension/src/background/GuardianSigner.ts
simonheys: Changed so you can optionally inject the cosigner function (via constructor) which receives the message, then it can be whatever one likes

packages/extension/src/background/accountMessaging.ts
simonheys: Hah yes indeed - there is already a `type` field too - needs a separate refactor, can probably re-use the existing shapes and types from transaction review, will open a ticket for this

packages/extension/src/background/wallet.ts
simonheys: Refactored this, but it's still inside the same class as it's all quite tightly coupled in there. See what you think.

packages/extension/src/background/shieldMessaging.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export const handleShieldMessage: HandleMessage<ShieldMessage> = async ({
+  msg,
+  background: { wallet },
+}) => {
+  switch (msg.type) {
+    case "SHIELD_MAYBE_ADD_ACCOUNT": {
+      if (!ARGENT_SHIELD_ENABLED) {
+        /** should never happen */
+        throw "Argent Shield is not enabled"
```
janek26, line 21: error

packages/extension/src/background/wallet.ts
```diff
@@ -293,9 +296,18 @@ export class Wallet {
 
     try {
       const accountWithTypes = await getAccountTypesFromChain(accounts)
-      return accountWithTypes
+      if (!ARGENT_SHIELD_ENABLED) {
+        return accountWithTypes
+      }
+      const accountWithTypesAndGuardians = await getAccountGuardiansFromChain(
```
janek26, line 302: lets use Promise.all to do all requests at the same time, so the multicall lib can do its job

simonheys: Refactored this, should also make it easer to add additional checks in future

packages/extension/src/ui/components/ControlledPinInput.tsx
```diff
@@ -0,0 +1,31 @@
+import { PinInput, PinInputProps } from "@chakra-ui/react"
```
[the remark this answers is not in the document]
simonheys: There is a follow-up 'UI' story, will make sure to do this then as there will be some styling changes too.

packages/extension/src/shared/account/update.ts
```diff
@@ -19,6 +21,12 @@ export async function updateAccountDetails(
     newAccounts = await getAccountTypesFromChain(allAccounts)
   }
 
+  if (ARGENT_SHIELD_ENABLED) {
+    if (scope === "guardian" || scope === "all") {
+      newAccounts = await getAccountGuardiansFromChain(allAccounts)
```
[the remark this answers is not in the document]
simonheys: I think this is done now, same as Promise.all change earlier

janek26 (approved):
thanks!

PR #1657, opened 2023-01-23
chore: extract guardian logic in its own package

bluecco:

Move guardian logic in its own package

### Changes

Moved all the logic and files used by guardian in `guardian` package

### Checklist

- [x] Code self-reviewed
- [x] Code self-tested


bluecco:
@janek26 @simonheys this is how I've proceeded in order to have guardian in its own package

@simonheys maybe we can check together how to test that everything is ok

packages/guardian/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@argent/guardian",
+  "version": "6.0.1",
```
simonheys, line 3:
```suggestion
  "version": "5.2.0",
```

packages/extension/package.json
```diff
@@ -80,6 +80,7 @@
     "@argent/x-multicall": "^5.2.0",
     "@argent/x-sessions": "^5.2.0",
     "@argent/x-swap": "^5.2.0",
+    "@argent/guardian": "^6.0.1",
```
simonheys, line 83:
```suggestion
    "@argent/guardian": "^5.2.0",
```

packages/guardian/.gitignore
```diff
@@ -0,0 +1,24 @@
+# Logs
```
simonheys, line 1: This whole file can be removed unless its specifying something other than what's in the root `.gitignore`?

bluecco: useless, will remove

packages/guardian/package.json
bluecco: right, I was starting from web wallet at the beginning

packages/guardian/package.json
```diff
@@ -0,0 +1,41 @@
...
+  "devDependencies": {
+    "@tsconfig/svelte": "^3.0.0",
+    "@types/async-retry": "^1.4.5",
+    "starknet": "^4.19.2",
+    "typescript": "^4.9.3",
+    "vite": "^3.0.0",
+    "vite-plugin-dts": "^1.4.0"
+  },
+  "peerDependencies": {
+    "starknet": "^4.19.2"
```
simonheys, line 39: This is specified in both dev and peer -  should be one or the other?

bluecco: will remove get-starknet-core, not used

packages/guardian/src/services/account.ts
```diff
@@ -0,0 +1,35 @@
+import { getJwt } from "./jwt"
+
+export const ARGENT_API_BASE_URL = process.env.ARGENT_API_BASE_URL as string
```
simonheys, line 3: I feel that `ARGENT_API_BASE_URL` and `cosignerSign` implementations are specific to the extension. e.g. the original fetcher method of cosignerSign also includes some version headers that are expected.

How about moving the types `Cosigner, CosignerMessage` into `GuardianSigner.ts` where they are referenced, then move the actual `cosignerSign` implementation back into the extension?

bluecco: works for me, I can always change that later if needed to be shared

packages/guardian/src/services/idb.ts
```diff
@@ -1,17 +1,26 @@
-import Dexie, { Table } from "dexie"
```
janek26, line 1: I'm not sure if we should keep this in shared, or if we should move everything storage out of this shared package, and just export a StorageInterface ?

simonheys: I agree to move it out of shared, the use case seems slightly different.

bluecco: 👍

packages/guardian/src/services/account.ts
bluecco: also @simonheys , for ARGENT_API_BASE_URL, I'll add the url in `cosignerSign`

resulting into `export const cosignerSign: Cosigner = async (url: string, message: CosignerMessage) => {...}`

packages/guardian/src/services/idb.ts
simonheys: Works well!

packages/guardian/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "build": "vite build",
+    "dev": "vite build --watch",
+    "prepare": "yarn build"
+  },
+  "dependencies": {
+    "jose": "^4.11.2",
+    "lodash-es": "^4.17.21"
+  },
+  "devDependencies": {
+    "@tsconfig/svelte": "^3.0.0",
```
simonheys, line 29: This package is not using Svelte? Could simplify by removing this and maybe using same tsconfig from similar package - maybe multicall package?

simonheys (changes requested):
Works great - just one final comment on config then good to merge

simonheys (approved):

PR #1672, opened 2023-01-31
feat: useNx

dhruvkelawala:

Speed up Lerna 

### Changes

Added Nx

### Checklist

- [x] Code self-reviewed
- [x] Code self-tested


janek26:
https://blog.nrwl.io/lerna-reborn-whats-new-in-v6-10aec6e9091c

I think most of this is enabled by default

dhruvkelawala:
https://blog.nrwl.io/lerna-reborn-whats-new-in-v6-10aec6e9091c#:~:text=flag%20to%20false.-,To%20experience,-fast%20caching%2C%20ensure

juristr:
Hey. So I've seen that you already have the latest Lerna version installed. What you might want to do though is to enable the caching capabilities. The https://lerna.js.org/upgrade has instructions and videos that demo how to set that up. But basically running `npx lerna add-caching` should help.

We also have a `#lerna` channel in the community slack where you can reach out with questions. Here's the link to join: https://go.nrwl.io/join-slack

As a side-note: once you have that, you can also enable remote caching for CI with [Nx Cloud](https://nx.app). Since you're an open-source project this would be completely free. Just ping us on our support email there (at the footer of the page) and we can enable it if you're interested.

dhruvkelawala:
@juristr I already have added `nx.json` in this PR using `npx lerna add-caching`

nx.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "nx/presets/npm.json",
+  "tasksRunnerOptions": {
+    "default": {
+      "runner": "nx/tasks-runners/default",
+      "options": {
+        "cacheableOperations": ["build", "lint", "test", "test:e2e"],
```
juristr, line 7: for `test:e2e` you need to make sure the operation is actually cacheable & idempotent. Meaning it doesn't have any side-effects, like calling a API etc which would be outside the controllable environment of Lerna and thus Lerna wouldn't be able to cache it properly

juristr:
Other than that it looks good now I think. As mentioned, the next step could be to activate remote caching, but I would do that in a separate step :)

lerna.json
```diff
@@ -2,5 +2,6 @@
   "packages": ["packages/*"],
   "version": "5.2.0",
   "npmClient": "yarn",
-  "useWorkspaces": true
+  "useWorkspaces": true,
+  "useNx": true
```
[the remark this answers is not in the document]
dhruvkelawala: `lerna repair` should have fixed it. Not sure why it was left there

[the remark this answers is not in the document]
dhruvkelawala:
```suggestion
```

nx.json
dhruvkelawala: Makes sense. Should remove e2e

dhruvkelawala:
```suggestion
        "cacheableOperations": ["build", "lint", "test"],
```

canlopes (approved):

diegodelrieu (approved):

PR #1675, opened 2023-01-31
fix: multisig feature flag

dhruvkelawala:

Fixes feature flag in NewAccountScreen


dhruvkelawala:
Missed a conditional to check for `enabled` even when it was already present

simonheys:
Now it's hidden, but it doesn’t show if I set `FEATURE_MULTISIG=true` - wondered if `const value = process.env[name]` was the culprit

dhruvkelawala:
Fixed

dhruvkelawala:
Copilot suggested me to use `process.env[name]`. Guess it's not correct

simonheys:
> Guess it's not correct always

Copilot = Subtle bug generator :)

simonheys (approved):

PR #1692, opened 2023-02-02
Feature/better-multicall

janek26:

Allows for an user to upgrade his account to a different account implementation.

### Changes

- add new screen to pick implementation
- rename setting "use plugin account" to "show implementation picker"
- make upgrade logic more robust

### Checklist

- [x] Code self-reviewed
- [x] Code self-tested


packages/extension/src/background/wallet.ts
```diff
@@ -201,13 +202,11 @@ export class Wallet {
     accountType: ArgentAccountType,
   ): Promise<string> {
```
diegodelrieu, line 203: type can be updated to `Promise<keyof Required<Network>["accountClassHash"]>` I guess no ? Or TS might be able to infer it

packages/extension/src/shared/network/type.ts
```diff
@@ -10,6 +10,7 @@ export interface Network {
   accountClassHash?: {
     argentAccount: string
     argentPluginAccount?: string
+    argentBetterMulticallAccount?: string
```
diegodelrieu, line 13: why better?

packages/extension/src/ui/components/Menu.tsx
```diff
@@ -0,0 +1,55 @@
+import styled from "styled-components"
```
diegodelrieu, line 1: Is there a particular reason to mix up Chakra and styled components ?

packages/extension/src/ui/services/addresses.ts
```diff
@@ -69,8 +69,11 @@ export const addressSchema = yup
 export const isValidAddress = (address: string) =>
   addressSchema.isValidSync(address)
 
-export const isEqualAddress = (a: string, b: string) => {
+export const isEqualAddress = (a: string, b?: string) => {
```
diegodelrieu, line 72: isnt this worth creating a new util ? it seems like it's serving a slightly different purpose here

packages/extension/src/shared/network/type.ts
janek26: https://community.starknet.io/t/proposal-to-improve-multicall-allowing-to-chain-transactions/2038

packages/extension/src/ui/services/addresses.ts
janek26: I think having it here is just fine, and undefined can not be equal to the address in string, so I think it keeps the functionality it promises

packages/extension/src/background/wallet.ts
janek26: I did not touch that code
While I think it's a good improvement, it has implications in the code where this function is used that I dont want to mess with right now for this feature

packages/extension/src/ui/features/accountEdit/AccountImplementationScreen.tsx
```diff
@@ -0,0 +1,148 @@
...
+        {otherImplementations && (
+          <>
+            <H3>Other implementations</H3>
+            <AutoColumn gap="md">
+              {otherImplementations.map((i) => (
+                <ImplementationItem
+                  {...i}
+                  key={i.id}
+                  active={false}
+                  onClick={async () => {
```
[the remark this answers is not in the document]
janek26: thanks :)

packages/extension/src/ui/components/Menu.tsx
janek26: no - we used styled-components before and are transitioning to chakra.

I just took existing code and moved it to be able to delete a much bigger file :)

packages/extension/src/ui/services/addresses.ts
dhruvkelawala: Do you think we can do this?

https://github.com/argentlabs/argent-x/blob/ea9b1c1eb412c7417de1ad0cbaed2efd4084abd8/packages/extension/src/ui/services/addresses.ts#L72

dhruvkelawala (approved):
lgtm

PR #1700, opened 2023-02-03
[BLO-597] enable deploy for local wallet

diegodelrieu:

`Dev should be able to deploy wallet in local network by sending one gwei to themself`

Screen Recording 2023-02-03 at 16.37.55.mov


packages/extension/src/ui/features/accountEdit/AccountEditScreen.tsx
```diff
@@ -37,7 +42,16 @@ import {
...
+const {
+  ExpandIcon,
+  HideIcon,
+  PluginIcon,
+  AlertIcon,
+  ArgentShieldIcon,
+  NetworkIcon,
+} = icons
+
+const LOCAL_NETWORK = "Localhost 5050"
```
janek26, line 54: lets not use the name (can be changed) but the id `localhost` please

packages/extension/src/ui/features/accountEdit/AccountEditScreen.tsx
```diff
@@ -107,6 +121,22 @@ export const AccountEditScreen: FC = () => {
       } Argent Shield…`
     : `Two-factor account protection`
 
+  const handleDeploy = () => {
+    if (account) {
+      const ONE_GWEI = getUint256CalldataFromBN(parseAmount("1", 0))
+      const self = account.address
+      const ETH_TOKEN_ADDRESS =
+        "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7"
```
janek26, line 129: Allow for custom feeTokens, use `getFeeToken(currentNetwork.id).address`

janek26 (approved):

packages/extension/src/ui/features/accountEdit/AccountEditScreen.tsx
diegodelrieu: ah right fees can be paid in things other than eth 🤯

janek26: not yet, but we are getting there :D

packages/extension/src/ui/features/accountEdit/AccountEditScreen.tsx
```diff
@@ -202,6 +230,11 @@ export const AccountEditScreen: FC = () => {
               Use Plugins
             </ButtonCell>
           )}
+          {account?.needsDeploy && (
```
bluecco, line 233: seems the check on localhost is gone?
from previous comment shouldn't it use `localhost` for this case?

diegodelrieu: Yes it is @janek26 suggested that we remove the condition and have that option for all networks

bluecco: I've understood the change was to use the id `localhost` instead of the name (but keep the condition).

ofc I could just have misunderstood that, let's wait for @janek26

janek26: both fine for me, I think showing it for every account is also good :)

bluecco (approved):

PR #1701, opened 2023-02-03
fix error being thrown on all recoveries

diegodelrieu:

Just a follow up fix of a previous PR. The error was being thrown for all seedphrase recoveries, even valid ones.


bluecco (approved):

janek26 (approved):
ups

PR #1703, opened 2023-02-06
[BLO-664] add zod to validate aspect api returns

diegodelrieu:

`Sanitize Aspect API responses with zod`

Some important changes: 
- Added MSW as the mocking library as is recommended by the vitest documentation https://vitest.dev/guide/mocking.html#requests 
- This means that we currently have two mocking libraries, MSW and vi-fetch, the latter is a dead library with 40 stars and no activity in the past year. MSW has over 11k stars and is in active dev. I[ have created a ticket](https://argent.atlassian.net/browse/BLO-734?atlOrigin=eyJpIjoiOTc2ZjA1NTdlNjdiNDJmMWE3OGMwZWQwNThiNzZkZTMiLCJwIjoiaiJ9) to phase out vi-fetch and replace it with MSW. Luckily, there's only a few instances of it.

### Changes

`Add zod library and implemented for Aspect API`


packages/extension/src/ui/features/accountNfts/aspect.service.ts
```diff
@@ -96,7 +90,7 @@ export const fetchAspectCollection = async (
 
   const url = join(baseUrl, `?${params}`, "&limit=50")
   const assets = await fetchNextAspectCollection(url, account, contractAddress)
-  if (Array.isArray(assets) && assets.length > 0) {
+  if (assets.length > 0) {
```
simonheys, line 93: Is this guaranteed to be an array now? If not we should keep the check as sometimes the API may return unexpected stuffs

diegodelrieu: yes it is zod will throw if it's not an array, but mb this is too strict ?

packages/extension/package.json
```diff
@@ -96,6 +97,7 @@
     "dexie": "^3.2.2",
     "dexie-react-hooks": "^1.1.1",
     "ethers": "^5.5.1",
+    "global": "^4.4.0",
```
janek26, line 100: can we remove this?

janek26 (approved):

packages/extension/package.json
diegodelrieu: ah yes my mistake

simonheys (approved):

PR #1708, opened 2023-02-07
[BLO-664] Apply estimated maxFee when sending a transaction

diegodelrieu:

`We are basically memoizing the fee computation so that the same fee is applied in the estimate screen and in the transaction execution itself`

### DIFF
--- a/packages/extension/src/background/transactions/fees/store.ts
+++ b/packages/extension/src/background/transactions/fees/store.ts
@@ -23,7 +23,6 @@ export const estimatedFeesStore = new ArrayStorage<EstimatedFeesEnriched>([], {
 const timestampInSeconds = (): number => Math.floor(Date.now() / 1000)
 
 export const addEstimatedFees = (estimatedFees: EstimatedFees) => {
-  console.log("calling addesimtated", estimatedFees)
   const newEstimatedFees = {
     ...estimatedFees,
     timestamp: timestampInSeconds(),
--- a/packages/extension/src/background/transactions/fees/store.ts
+++ b/packages/extension/src/background/transactions/fees/store.ts
@@ -1,0 +1,50 @@
+import { isEqual } from "lodash-es"
+import { Call } from "starknet5"
+
+import { ArrayStorage } from "../../../shared/storage"
+
+type EstimatedFees = {
+  amount: string
+  suggestedMaxFee: string
+  accountDeploymentFee?: string
+  maxADFee?: string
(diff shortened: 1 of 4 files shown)


packages/extension/src/background/transactions/fees/store.ts
```diff
@@ -0,0 +1,49 @@
...
+  transactions: Call | Call[]
+}
+
+type EstimatedFeesEnriched = EstimatedFees & {
+  timestamp: number
+}
+
+export const estimatedFeesStore = new ArrayStorage<EstimatedFeesEnriched>([], {
+  namespace: "core:estimatedFees",
+  areaName: "local",
```
janek26, line 20: lets use "session" to store in memory only

packages/extension/src/background/transactions/transactionMessaging.ts
```diff
@@ -32,6 +33,14 @@ export const handleTransactionMessage: HandleMessage<
       const selectedAccount = await wallet.getSelectedAccount()
       const starknetAccount = await wallet.getSelectedStarknetAccount()
       const transactions = msg.data
+      const preComputedFees = await getEstimatedFees(transactions)
+
+      if (preComputedFees) {
```
janek26, line 38: I would keep the constant recalculation for ESTIMATE_TRANSACTION_FEE honestly.
And store the value every time.

Then consume the value only in the action where the transaction gets actually send. 

Can we also add analytics to the execute, how many times a transaction was send with cached value, and how many times it's missing that value? That way we can monitor adoption and catch bugs early

packages/extension/src/background/transactions/transactionExecution.ts
```diff
@@ -52,6 +53,11 @@ export const executeTransactionAction = async (
 ) => {
   const { transactions, abis, transactionsDetail, meta = {} } = action.payload
   const allTransactions = await transactionsStore.get()
+  const preComputedFees = await getEstimatedFees(transactions)
+
+  analytics.track("executeTransaction", {
```
bluecco, line 58: it this tracking useful?
maybe track values instead of just a boolean value?

diegodelrieu: So the point of the tracking was mostly to see whether the feature was working as expected and how often we are using the cached data

bluecco (approved):

janek26 (approved):

PR #1714, opened 2023-02-08
feat: NFT Support in signing screen

dhruvkelawala:

Support NFTs

Screenshot 2023-02-08 at 14 06 33

### Checklist

- [ ] Rebased to the last commit of the target branch (or merged)
- [ ] Code self-reviewed
- [ ] Code self-tested
- [ ] Tests updated (if needed)
- [ ] All tests are passing locally

### DIFF
--- a/packages/extension/src/ui/features/actions/transaction/useTransactionSimulatedData.ts
+++ b/packages/extension/src/ui/features/actions/transaction/useTransactionSimulatedData.ts
@@ -9,6 +9,8 @@ import {
   TokenDetails,
   TransactionSimulationTransfer,
 } from "../../../../shared/transactionSimulation/types"
+import { parseIPFSUri } from "../../../services/ipfs"
+import { useAspectContractAddresses } from "./../../accountNfts/aspect.service"
 import { Account } from "../../accounts/Account"
 import { useSelectedAccount } from "../../accounts/accounts.state"
 import { useTokensRecord } from "../../accountTokens/tokens.state"
@@ -31,8 +33,13 @@ interface Recipient {
   usdValue?: number
 }
 
-interface TransferSimulationData extends CommonSimulationData {
-  token: Token
+export interface TokenWithType extends Token {
+  type: "erc20" | "erc721" | "erc1155"
+  tokenId?: string
+}
+
+export interface AggregatedSimData extends CommonSimulationData {
+  token: TokenWithType
   recipients: Recipient[]
   approvals: ApprovalSimulationData[]
   safe?: boolean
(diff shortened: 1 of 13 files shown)


packages/extension/src/ui/features/actions/transaction/useTransactionSimulatedData.ts
```diff
@@ -58,60 +85,106 @@ export const useAggregatedSimData = (
...
 
         const usdValue = transfers.reduce<number>((acc, t) => {
-          if (!t.details?.usdValue) {
+          if (!t.usdValue) {
             return acc
           }
 
           if (t.from === account?.address) {
-            return acc - parseFloat(t.details.usdValue)
+            return acc - parseFloat(t.usdValue)
```
simonheys, line 184: I get nervous about float and it's flaky precision for accumulating currency values if that's what's happening here? - one alternative I think `bignumber.js` (the one that supports decimal numbers) is used for this purpose elsewhere

packages/extension/src/ui/services/ipfs.ts
```diff
@@ -0,0 +1,5 @@
+export function parseIPFSUri(uri: string): string {
```
simonheys, line 1: Is this actually used anywhere yet?

dhruvkelawala: Nope, should remove it

packages/extension/src/ui/features/actions/transaction/useTransactionSimulatedData.ts
dhruvkelawala: Right!

diegodelrieu (approved):

simonheys (approved):

bluecco (approved):

diegodelrieu:
Seems to work fine locally, just one minor detail, when I have several NFTs I dont think they're in the correct order here 

I think we would expect NFT > NFT .... > Total

image

diegodelrieu:
> 

I mean here it's literally the total from my cart in Aspect. I get that conceptually this is not exactly the same as an ecommerce but from a user PoV it is. I think at least grouping the outgoing and the incoming would make sense.

janek26:
Up to Ismael if he would like to order the events, but I think having all incoming transfers first, and then all outgoing would be nice. Within these categories it would be nice to have all NFTs first and then all ERC20.

But happy to have a different ticket

dhruvkelawala:
Screenshot 2023-02-14 at 14 21 49

Good enough for now?

PR #1732, opened 2023-02-09
[BLO-729] Activity view should display fee cost

diegodelrieu:

`There were already network fees but they were purely relying on backend data. Now there's a fallback with starknet js that way the fees should always be visible for a completed transaction`

### Changes

Added a custom hook to get the fees + test on it

image

### DIFF
--- a/packages/extension/src/ui/features/accountActivity/useTransactionFees.ts
+++ b/packages/extension/src/ui/features/accountActivity/useTransactionFees.ts
@@ -1,0 +1,24 @@
+import { useEffect, useState } from "react"
+
+import { Network, getProvider } from "../../../shared/network"
+import { TransformedTransaction } from "./transform/type"
+
+export const useTransactionFees = ({
+  network,
+  transactionTransformed,
+  hash,
+}: {
+  network: Network
+  transactionTransformed: TransformedTransaction
+  hash?: string
+}) => {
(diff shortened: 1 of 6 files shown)


packages/extension/src/ui/features/accountActivity/useTransactionFees.ts
```diff
@@ -0,0 +1,24 @@
...
+}) => {
+  const [txFee, setTxFee] = useState<string | undefined>()
+  useEffect(() => {
+    const getFeeFromStarknetJs = async () => {
+      const receipt = await getProvider(network).getTransactionReceipt(hash)
+      setTxFee(transactionTransformed.actualFee ?? receipt.actual_fee)
+    }
+    getFeeFromStarknetJs()
+  }, [hash, network, transactionTransformed.actualFee])
+  return txFee
```
simonheys, line 23: `swr` is used quite a lot elsewhere in codebase for simplifying this pattern; keeps state and calls the fetcher when the deps change, pseudocode for example;

`const { data: txFee } = useSWR([hash, network, transactionTransformed.actualFee], getFeeFromStarknetJs)`

if the returned value will never change then can also consider using swr/immutable

simonheys (approved):

bluecco (approved):

PR #1741, opened 2023-02-13
[BLO-768] Transaction simulation Send and Declare Contract screens

diegodelrieu:

`Implemented the approval screen for declare contract and send transactions`

#### Declare contract
image

#### Send transaction
image

### Changes

- Reorganized a lot of the files (if a component is imported only in one place we group them in a folder)
- Split them into more digestible chunks 
- Wrote some basic unit tests (only on `TransactionIcon` because the PR was already getting quite long)

### Checklist

- [x] Code self-reviewed
- [x] Code self-tested


packages/extension/src/ui/features/actions/transaction/ApproveTransactionScreen/DappHeader/TransactionIcon/IconWrapper.tsx
```diff
@@ -0,0 +1,23 @@
+import { Center } from "@chakra-ui/react"
+import { BoxProps } from "@mui/material"
```
simonheys, line 2: Should not use Mui any more - instead of `BoxProps` there should be `CenterProps` from Chakra

packages/extension/src/ui/features/actions/transaction/ApproveTransactionScreen/DappHeader/TransactionIcon/SendTransactionIcon.tsx
```diff
@@ -0,0 +1,68 @@
...
+        <Center
+          w="14"
+          h="14"
+          background="neutrals.600"
+          borderRadius="90"
+          boxShadow="menu"
+          padding="4"
+        >
+          <SendIcon fontSize={"4xl"} color="white" />
+          {/* // what's the fallback token image ?  */}
```
simonheys, line 34:
```suggestion
          {/* // what's the fallback token image ?  */}
```

packages/extension/src/shared/transactionReview.service.ts
```diff
@@ -44,12 +44,20 @@ export interface ApiTransactionReviewResponse {
...
+  transfer: "transfer",
+} as const
+
 export type ApiTransactionReviewActivityType =
-  | "account-upgrade"
-  | "approve"
-  | "set-approval-for-all"
-  | "swap"
-  | "transfer"
+  (typeof apiTransactionReviewActivityType)[keyof typeof apiTransactionReviewActivityType]
```
janek26, line 56: can we do this instead? I think it's simpler to understand and less data duplication
```
export const apiTransactionReviewActivityType = [
  "account-upgrade",
  "approve",
  "set-approval-for-all",
  "swap",
  "transfer",
] as const

export type ApiTransactionReviewActivityType =
  (typeof apiTransactionReviewActivityType)[number]
```

packages/extension/src/shared/transactionReview.service.ts
```diff
@@ -212,3 +220,22 @@ export const getTransactionReviewHasSwap = (
...
+export const getTransactionReviewWithType = (
+  transactionReview?: ApiTransactionReviewResponse,
+): TransactionReviewWithType | undefined => {
+  if (!transactionReview) {
+    return
+  }
+  for (const review of transactionReview.reviews) {
+    if (
+      review.activity?.type &&
+      apiTransactionReviewActivityType[review.activity?.type]
```
janek26, line 233: `review.activity?.type` is already of type `ApiTransactionReviewActivityType`. Why do you double check this?

If still needed this could look like
```suggestion
      apiTransactionReviewActivityType.includes(review.activity.type)
```

packages/extension/src/ui/features/actions/transaction/ApproveTransactionScreen/DappHeader/TransactionIcon/DeclareTransactionIcon.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+const { DocumentIcon } = icons
+export const DeclareContractIcon = () => {
+  return (
+    <IconWrapper
+      data-testid="declare-contract-icon"
+      position="relative"
+      borderRadius="90"
+    >
+      <DocumentIcon />
```
simonheys, line 13: This icon looks too small, it's 24px in designs so adding `fontSize="2xl"` should fix it here (I haven't tried it though)

packages/extension/src/shared/transactionReview.service.ts
diegodelrieu: it resolves to the same type but the issue with the `[] as const` is that you can't use it as an enum, for example in the tests

diegodelrieu: I find it reads much better to do `apiTransactionReviewActivityType.swap` than `apiTransactionReviewActivityType[apiTransactionReviewActivityType.indexOf('swap')]`

janek26: you map the same key to the same value, so why would you ever do something like `apiTransactionReviewActivityType[apiTransactionReviewActivityType.indexOf('swap')]`?
and not just `swap` directly.

simonheys (approved):

PR #1743, opened 2023-02-14
feat: order tx sim results

janek26:


packages/extension/src/ui/features/actions/transaction/useTransactionSimulatedData.ts
```diff
@@ -65,6 +65,41 @@ export type ValidatedTokenApproval = Omit<
...
+  const orderedSimData = orderBy(simData, (t) => t.amount.toString(16), "desc")
+
+  const [erc721Transfers, restTransfers] = partition(
+    orderedSimData,
+    (t) => t.token.type === "erc721",
+  )
+  const [erc1155Transfers, erc20Transfers] = partition(
+    restTransfers,
+    (t) => t.token.type === "erc1155",
+  )
```
dhruvkelawala, line 84:
```suggestion
  const [erc20Transfers, erc1155Transfers] = partition(
    restTransfers,
    (t) => t.token.type === "erc20",
  )
```

No support for erc1155 for now. Better to do it this way

dhruvkelawala (changes requested):

packages/extension/src/ui/features/actions/transaction/useTransactionSimulatedData.ts
```diff
@@ -65,6 +65,41 @@ export type ValidatedTokenApproval = Omit<
...
 }
 
+function partitionIncomingOutgoingTransfers(transfers: AggregatedSimData[]) {
+  return partition(transfers, (t) => t.amount.isGreaterThan(0))
+}
+
+function orderAggregatedSimData(
+  simData: AggregatedSimData[],
+): AggregatedSimData[] {
+  const orderedSimData = orderBy(simData, (t) => t.amount.toString(16), "desc")
```
dhruvkelawala, line 75:
```suggestion
```

Not needed now

dhruvkelawala (changes requested):

packages/extension/src/ui/features/actions/transaction/useTransactionSimulatedData.ts
dhruvkelawala: Sure, I feel erc20 partitioning makes more sense but sure your works too.

packages/extension/src/ui/features/actions/transaction/useTransactionSimulatedData.ts
janek26: partitioning is stable, so this will still allow for the groups to be sorted in itself, which is nice to have, no? :)

packages/extension/src/ui/features/actions/transaction/useTransactionSimulatedData.ts
janek26: I think you're right, lets have 
1. erc721 transfers
2. all other transfers (including 1155)
3. erc20 transfers

packages/extension/.eslintrc.js
```diff
@@ -40,5 +40,17 @@ module.exports = {
     ],
     "@typescript-eslint/no-non-null-assertion": "error",
     curly: "error",
+    "@typescript-eslint/no-restricted-imports": [
```
diegodelrieu, line 43: shouldnt we just get rid of lodash ?

simonheys (approved):

packages/extension/.eslintrc.js
janek26: that's what this eslint rule will do, no?

It's no dependency to the project, still it wont complain if you use it, as it's a nested dependency for some dependency we use.

PR #1768, opened 2023-02-16
BLO-739 fix: 2fa state transition

simonheys:

Fixes an issue where the 'live' state of 2FA on account was not in sync with on-chain state and would show momentarily in the wrong state

### Changes

- add a hook to react to both tx and on-chain state, resolving when on-chain state is updated
- also display tx failure

### Checklist

- [x] Code self-reviewed
- [x] Code self-tested

### DIFF
--- a/packages/extension/src/ui/features/accountEdit/AccountEditScreen.tsx
+++ b/packages/extension/src/ui/features/accountEdit/AccountEditScreen.tsx
@@ -9,7 +9,7 @@ import {
   icons,
 } from "@argent/ui"
 import { Center, Flex, Image, Spinner } from "@chakra-ui/react"
-import { FC, useCallback, useState } from "react"
+import { FC, useCallback, useMemo, useState } from "react"
 import { Link, useNavigate, useParams } from "react-router-dom"
 
 import { settingsStore } from "../../../shared/settings"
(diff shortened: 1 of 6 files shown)


bluecco (approved):

packages/extension/src/ui/features/accountEdit/AccountEditScreen.tsx
```diff
@@ -102,11 +102,22 @@ export const AccountEditScreen: FC = () => {
...
+        ? "Adding Argent Shield Failed"
+        : "Removing Argent Shield Failed"
+    }
+    if (status === "PENDING") {
+      return isAdding ? "Adding Argent Shield…" : "Removing Argent Shield…"
+    }
+    return "Two-factor account protection"
+  }, [isAdding, status])
+
+  const shieldIsLoading = liveAccountGuardianState.status === "PENDING"
```
diegodelrieu, line 120: this should be in a usememo no ?

packages/extension/src/ui/features/shield/ShieldBaseFinishScreen.tsx
```diff
@@ -13,76 +16,94 @@ const {
...
+      icon: AlertIcon,
+      title: isAdding
+        ? "Adding Argent Shield Failed"
+        : "Removing Argent Shield Failed",
+      subtitle: `${accountName} was not modified because the transaction failed. Please try again later`,
+      variant: "danger",
+    }
+  }
+  if (status === "PENDING") {
+    return {
```
diegodelrieu, line 49: mb instead of having ternaries everywhere we can have just one if ?

diegodelrieu (approved):

packages/extension/src/ui/features/accountEdit/AccountEditScreen.tsx
simonheys: I believe is okay if it's not a complex / slow operation, memoizing and checking the memo will make it slower

PR #1774, opened 2023-02-17
fix: feature flag

simonheys:

Unused feature flag

### Changes

- remove unused feature flag

### Checklist

- [x] Code self-reviewed
- [x] Code self-tested


packages/extension/src/ui/features/actions/transaction/useTransactionSimulation.ts
```diff
@@ -12,6 +12,10 @@ import { ApiTransactionSimulationResponse } from "../../../../shared/transaction
 import { argentApiFetcher } from "../../../services/argentApiFetcher"
 import { useConditionallyEnabledSWR } from "../../../services/swr"
 import { Account } from "../../accounts/Account"
+import { useIsMainnet } from "../../networks/useNetworks"
+
+export const isTransactionSimulationEnabled =
+  (process.env.FEATURE_TRANSACTION_SIMULATION || "false") === "true"
```
janek26, line 18: should we just remove the feature flag?

All of this is unuseable without simulation, that's why we added the fallback simulation (which would be disabled by this flag too) 

I think we should just remove it

simonheys: Ah... I misunderstood where the fallback was, have seen it now and will rework this

simonheys: Reworked and removed flag

packages/extension/src/shared/transactionSimulation/transactionSimulation.service.ts
```diff
@@ -10,6 +10,8 @@ import type {
   TransactionSimulationTransfer,
 } from "./types"
 
+const supportedBackendSimulationChainIds = [constants.StarknetChainId.MAINNET]
```
janek26, line 13: why should we introduce this?

If the simulation fails for testnet (backend not ready) it will fallback to direct simulation. As soon as backend comes online this should work.

simonheys: It fails with a 500 at the moment, causing throttling

janek26 (approved):
😂 sorry for making this PR so small

diegodelrieu (approved):

PR #1780, opened 2023-02-20
[BLO-790] User can fill in a multisg creation form

diegodelrieu:

Form for multisig creation (without backend call)


packages/extension/scripts/hot-reload-server.ts
```diff
@@ -1,7 +1,7 @@
 import path from "path"
 
 import chokidar from "chokidar"
-import { debounce } from "lodash-es"
+import { debounce } from "lodash"
```
dhruvkelawala, line 4: Why?

packages/extension/src/ui/features/accounts/multisig/CreateMultisigScreen/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+`
+
+const StyledPageWrapper = styled(PageWrapper)`
+  ${({ theme }) => theme.mediaMinWidth.md`
+    > ${Panel}:last-child {
+      background:url('./assets/onboarding-background.svg') no-repeat center;
+      background-size: cover;
+    }
+  `}
+`
```
dhruvkelawala, line 58: We are moving away from styled-components and sticking with chakra-ui & argent-ui for everything new

dhruvkelawala (changes requested):

dhruvkelawala:
Screenshot 2023-02-20 at 11 44 15

I think it's better to validate "incorrect signer" on submit (not onBlur), because user might be waiting for other signers and I don't think we should show error message at that time

packages/extension/src/ui/features/accounts/multisig/CreateMultisigScreen/index.tsx
diegodelrieu: Ok yeah this was just lazy copy pasta from a similar page but will migrate to chakra

dhruvkelawala (approved):

PR #1785, opened 2023-02-21
feat: optimistically increment nonce in session wallet

prof197:

Updated the session wallet to increment the account's nonce value optimistically in serial transactions. Without this, if the account submits a 2nd (3rd, etc) transaction before the 1st is `accepted on l2`, there is a nonce failure for all but the 1st.

### Changes

For session accounts only:
- maintain a list of "open transactions" that have been submitted (but would not yet have incremented the nonce returned by the `getNonceByAddress`)
- add that many transactions to the nonce passed into the `execute` function

### Checklist

- [x] Code self-reviewed
- [x] Code self-tested

### DIFF
--- a/packages/sessions/src/account.ts
+++ b/packages/sessions/src/account.ts
@@ -30,7 +30,7 @@ import {
 
 export class SessionAccount extends Account implements AccountInterface {
   public merkleTree: merkle.MerkleTree
-  private openTransactions: string[] = []
+  private openTransactionHashes: string[] = []
 
   constructor(
     providerOrOptions: ProviderOptions | ProviderInterface,
@@ -138,28 +138,32 @@ export class SessionAccount extends Account implements AccountInterface {
 
   // check for resolution of any transactions that are not yet counted by account's nonce
   async getOpenTransactionsCount() {
-    const receipts = await Promise.all(
-      this.openTransactions
-        .filter((txHash) => !!txHash)
-        .map((txHash) => this.getTransactionReceipt(txHash)),
-    )
-    receipts.forEach((receipt) => {
-      // if the the transaction status is not "pending" (or prior), transaction should have been counted by nonce
-      if (
-        receipt &&
-        !["NOT_RECEIVED", "RECEIVED", "PENDING"].includes(receipt.status)
-      ) {
-        const removeIndex = this.openTransactions.indexOf(
-          receipt.transaction_hash,
-        )
-        if (removeIndex > -1) {
-          this.openTransactions.splice(removeIndex, 1)
+    try {
+      const receipts = await Promise.all(
+        this.openTransactionHashes
+          .filter((txHash) => !!txHash)
+          .map((txHash) => this.getTransactionReceipt(txHash)),
+      )
+      receipts.forEach((receipt) => {
+        // if the the transaction status is not "pending" (or prior), transaction should have been counted by nonce
+        if (
+          receipt &&
+          !["NOT_RECEIVED", "RECEIVED", "PENDING"].includes(receipt.status)
+        ) {
+          const removeIndex = this.openTransactionHashes.indexOf(
+            receipt.transaction_hash,
+          )
+          if (removeIndex > -1) {
+            this.openTransactionHashes.splice(removeIndex, 1)
+          }
         }
-      }
-    })
+      })
 
-    // return the number of transactions not yet counted by nonce
-    return this.openTransactions.length
+      // return the number of transactions not yet counted by nonce
+      return this.openTransactionHashes.length
+    } catch {
+      return 0
+    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


packages/sessions/src/account.ts
```diff
@@ -135,6 +136,32 @@ export class SessionAccount extends Account implements AccountInterface {
     }
   }
 
+  // check for resolution of any transactions that are not yet counted by account's nonce
+  async getOpenTransactionsCount() {
+    const receipts = await Promise.all(
+      this.openTransactions
+        .filter((txHash) => !!txHash)
+        .map((txHash) => this.getTransactionReceipt(txHash)),
```
janek26, line 144: quite heavy method, especially if you only use status later on.
I think there is this.getTransactionStatus or sth

packages/sessions/src/account.ts
```diff
@@ -30,6 +30,7 @@ import {
 
 export class SessionAccount extends Account implements AccountInterface {
   public merkleTree: merkle.MerkleTree
+  private openTransactions: string[] = []
```
janek26, line 33:
```suggestion
  private openTransactionHashes: string[] = []
```

packages/sessions/src/account.ts
```diff
@@ -172,9 +199,13 @@ export class SessionAccount extends Account implements AccountInterface {
 
     const version = number.toBN(hash.transactionVersion)
 
+    const optimisticNonce = transactionsDetail.nonce
+      ? nonce
+      : number.toBN(parseInt(nonce) + (await this.getOpenTransactionsCount()))
```
janek26, line 204: please move result of `this.getOpenTransactionsCount` into own variable

packages/sessions/src/account.ts
```diff
@@ -135,6 +136,32 @@ export class SessionAccount extends Account implements AccountInterface {
     }
   }
 
+  // check for resolution of any transactions that are not yet counted by account's nonce
+  async getOpenTransactionsCount() {
+    const receipts = await Promise.all(
```
janek26, line 141: can we error handle this properly? Maybe use `Promise.allSettled` or add a try-catch in case the .all fails, and default to `return 0`

packages/sessions/src/account.ts
```diff
@@ -188,7 +219,8 @@ export class SessionAccount extends Account implements AccountInterface {
 
     const calldata = transaction.fromCallsToExecuteCalldata(transactions)
 
-    return this.invokeFunction(
+    this.openTransactions.push("")
```
janek26, line 222: why?

I think the array should contain only valid transaction hashes

packages/sessions/src/account.ts
```diff
@@ -172,9 +199,13 @@ export class SessionAccount extends Account implements AccountInterface {
 
     const version = number.toBN(hash.transactionVersion)
 
+    const optimisticNonce = transactionsDetail.nonce
+      ? nonce
```
janek26, line 203: this is already handled in line 184. Please add your logic there.

packages/sessions/src/account.ts
```diff
@@ -197,8 +229,21 @@ export class SessionAccount extends Account implements AccountInterface {
...
+        nonce: optimisticNonce,
       },
     )
+    invocation.then((response) => {
+      const removeIndex = this.openTransactions.indexOf("")
+      if (removeIndex > -1) {
+        this.openTransactions[removeIndex] = response.transaction_hash
+      }
+    })
+    invocation.catch(() => {
```
janek26, line 241: I think using try-catch with await is better than hooking into .then and .catch

janek26: error should be re-thrown too in that case

janek26 (changes requested):
Thanks for the PR! I found a couple of things that I think should be fixed before merge

packages/sessions/src/account.ts
prof197: `getTransactionStatus` is unfortunately only on `SequencerProvider`, but the AccountInterface the session account implements extends the base `Provider` which only has `getTransactionReceipt`

packages/sessions/src/account.ts
prof197: If starknet is being really slow to respond to `starknet_addInvokeTransaction`, then I think we would still want these empty placeholders to add to the length of openTransactionHashes (to add to the nonce optimistically). Otherwise, the user would still run into nonce conflicts. Usually (hopefully), this will be a pretty short state, but IMO it might be worth having for when the network is really slow.

packages/sessions/src/account.ts
prof197: I added some comments to clarify this some. This nonce may be different than the one calculated in line 184 (now 188 I think) because the non-optimistic nonce must be used for the estimateFee call (or it will fail), but the optimistic nonce must be calculated here.

prof197:
Thanks for the review! I made a couple comments where I'm not sure there is a great alternative, but I think I fixed everything else.

PR #1821, opened 2023-02-23
Blo 832

dhruvkelawala:

Stores additional multisig data (threshold, signers) in a different storage array while maintaining the original WalletAccount Array Store

### Changes

- add baseMultisigWalletStore
- 
### Checklist

- [ ] Rebased to the last commit of the target branch (or merged)
- [ ] Code self-reviewed
- [ ] Code self-tested
- [ ] Tests updated (if needed)
- [ ] All tests are passing locally


packages/extension/src/shared/api/constants.ts
```diff
@@ -2,6 +2,8 @@ import { isString } from "lodash-es"
 import urlJoin from "url-join"
 
 export const ARGENT_API_BASE_URL = process.env.ARGENT_API_BASE_URL as string
+export const ARGENT_MULTISIG_BASE_URL = process.env
```
diegodelrieu, line 5: probably should check that the variable is present

packages/extension/src/ui/features/accounts/multisig/multisig.service.ts
```diff
@@ -22,18 +22,18 @@ export const networkToStarknetNetwork = (network: Network) => {
   }
 }
 
+export interface ApiMultsigContent {
```
diegodelrieu, line 25: best to infer this from a zod schema to validate the response at the same time

packages/extension/src/ui/services/backgroundAccounts.ts
```diff
@@ -64,6 +65,32 @@ export const createNewMultisigAccount = async (
...
+    ])
+
+    if (res === "error") {
+      throw Error("Could not calculate multisig address")
+    }
+
+    return res
+  } catch (e) {
+    console.error(e)
+    throw Error("Could not add new account")
```
diegodelrieu, line 90: not sure about the double error handling here

packages/extension/src/shared/api/constants.ts
```diff
@@ -2,6 +2,8 @@ import { isString } from "lodash-es"
...
+  .ARGENT_MULTISIG_BASE_URL as string
```
dhruvkelawala, line 6: It's present here @diegodelrieu

packages/extension/src/ui/features/accounts/usePublicKey.ts
```diff
@@ -15,3 +20,46 @@ export const usePublicKey = (account?: BaseWalletAccount) => {
 
   return pubKey
 }
+
+export const useNextPublicKey = () => {
```
diegodelrieu, line 24: we can create a different file for this

packages/extension/src/background/wallet.ts
```diff
@@ -817,6 +885,55 @@ export class Wallet {
...
+    const multisigAccount = await getMultisigAccountFromBaseWallet(
+      baseMultisigAccount,
+    )
+
+    if (!multisigAccount) {
+      throw new Error("This multisig account does not exist")
+    }
+
+    const starkPair = await this.getKeyPairByDerivationPath(
+      multisigAccount.signer.derivationPath,
```
diegodelrieu, line 900: I'm slightly confused about the difference between the `signer` and the `signers`

packages/extension/src/shared/multisig/store.ts
```diff
@@ -0,0 +1,121 @@
...
+    .filter((account): account is MultisigWalletAccount => !!account) // If the account is hidden, it will be undefined and filtered out here
+}
+
+export async function getMultisigAccountFromBaseWallet(
+  baseWalletAccount: BaseWalletAccount,
+): Promise<MultisigWalletAccount | undefined> {
+  const baseMultisigWalletAccounts = await getBaseMultisigAccounts()
+  const walletAccounts = await getAccounts()
+
+  const bma = baseMultisigWalletAccounts.find((baseMultisigWalletAccount) =>
```
diegodelrieu, line 59: personally not a fan of the acronyms here

packages/extension/src/ui/features/accounts/GroupedAccountList.tsx
```diff
@@ -12,6 +13,7 @@ export interface GroupedAccountListProps {
   icon: React.ReactNode
   selectedAccount?: Account
   returnTo?: string
+  type: "standard" | "multisig"
```
diegodelrieu, line 16: this should probably come from a shared type and not be defined here

diegodelrieu: could have a `const accountType = { standard: "standard", multisig: "multisig"} as const` somewhere and just have `accountType.multisig` when we need it or the other way around. Or enums.

packages/extension/src/shared/multisig/store.ts
```diff
@@ -0,0 +1,121 @@
...
+        accountsEqual(walletAccount, baseMultisigAccount),
+      )
+
+      return {
+        ...walletAccount,
+        ...baseMultisigAccount,
+        type: "multisig",
+      }
+    })
+    .filter((account): account is MultisigWalletAccount => !!account) // If the account is hidden, it will be undefined and filtered out here
```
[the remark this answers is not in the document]
dhruvkelawala: The only time an account will be undefined is when it's present in `baseMultisigAccounts` store and not in `accounts` store. This case is possible only when an account is `hidden`. `hidden` attribute` is present in `WalletAccount` of `accounts` store. Does it make sense?

packages/extension/src/background/wallet.ts
dhruvkelawala: `signer` is the publicKey of current account and `signers` is the publicKeys of all signers of multisig? We pass signers in `constructorCalldata` of multisig deployment which affects the address. And we also use `signer` as addressSalt

packages/extension/src/ui/services/backgroundAccounts.ts
```diff
@@ -166,16 +198,28 @@ export const getPrivateKey = async () => {
...
   return publicKey
 }
 
+export const getNextPublicKey = async (networkId: string) => {
+  sendMessage({
+    type: "GET_NEXT_PUBLIC_KEY",
+    data: { networkId },
+  })
+
+  const { publicKey } = await waitForMessage("GET_NEXT_PUBLIC_KEY_RES")
```
diegodelrieu, line 218: can this fail ?

packages/extension/src/ui/features/accounts/usePublicKey.ts
```diff
@@ -14,3 +20,46 @@ export const usePublicKey = () => {
...
+  const [pubKey, setPubKey] = useState<string>()
+
+  const getNextPubKeyCallback = useCallback(
+    () => getNextPublicKey(network.id),
+    [network.id],
+  )
+
+  useEffect(() => {
+    getNextPubKeyCallback().then(setPubKey)
+    // on mount
```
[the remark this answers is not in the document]
dhruvkelawala: You mean a cleanup function? Isn't it the same as other `usePublicKey` and `useSigner` hooks?

packages/extension/src/shared/multisig/store.ts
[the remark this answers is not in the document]
diegodelrieu: I guess it makes some sense but I find it a bit convoluted, could we not filter based on hidden here instead ?

packages/extension/src/ui/features/accounts/GroupedAccountList.tsx
dhruvkelawala: https://github.com/argentlabs/argent-x/blob/56f5b15900ab11e33d5bf9cca9506bdbf4145922/packages/extension/src/shared/wallet.model.ts#L4

We already have it

packages/extension/src/ui/features/accounts/usePublicKey.ts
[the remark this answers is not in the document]
diegodelrieu: actually that's true. I wonder if we need to, I think we make extra requests by not cleaniing it up but could be wrong.

packages/extension/src/shared/multisig/store.ts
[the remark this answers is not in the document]
dhruvkelawala: https://github.com/argentlabs/argent-x/blob/3348fe9d5b27627c8563d17521ef1d8bb3dc5cc8/packages/extension/src/shared/multisig/store.ts#L33

Here we are using a selector, by default we are using `withoutHiddenSelector` which means, we won't fetch the accounts which are `hidden` in the accounts store. For those accounts, we will have `undefined` multisig accounts. Which just means that particular mutlsig account has been hidden by the user. There is also a `withHiddenSelector`

diegodelrieu (approved):

packages/extension/src/ui/features/multisig/Multisig.ts
```diff
@@ -0,0 +1,102 @@
...
+  BaseMultisigWalletAccount,
+  MultisigData,
+} from "../../../shared/wallet.model"
+import { createNewMultisigAccount } from "../../services/backgroundAccounts"
+import { Account, AccountConstructorProps } from "../accounts/Account"
+
+export interface MultisigConstructorProps extends AccountConstructorProps {
+  signers: string[]
+  threshold: number
+  creator?: string
```
simonheys, line 12: Add a comment about how this identifies 'creator' - is it some sort of key, or whatever?

simonheys (approved):

PR #1825, opened 2023-02-23
BLO-780 fix: remove appstate for managing errors in lock screen

bluecco:

`Tx error is displayed on login screen`

### Changes

- refactored lock screen and form to use local state instead of app state (had registered the tx error)
- tested locally with ionel too (to replicate, send max currency to another account and make another transaction while the first one is still pending)

### Checklist

- [x] Code self-reviewed
- [x] Code self-tested

### DIFF
--- a/packages/extension/src/ui/features/lock/LockScreen.tsx
+++ b/packages/extension/src/ui/features/lock/LockScreen.tsx
@@ -64,7 +64,7 @@ export const LockScreen: FC = () => {
                 navigate(target)
                 return true
               } catch {
-                setError("Invalid password")
+                setError("Incorrect password")
                 return false
               } finally {
                 setIsLoading(false)
--- a/packages/extension/src/ui/features/lock/LockScreen.tsx
+++ b/packages/extension/src/ui/features/lock/LockScreen.tsx
@@ -18,7 +18,7 @@ export const LockScreen: FC = () => {
   const navigate = useNavigate()
   const actions = useActions()
   const [isLoading, setIsLoading] = useState<boolean>(false)
-  const [error, setError] = useState<string | undefined>(undefined)
+  const [error, setError] = useState<string>()
   return (
     <Flex flex={1} flexDirection={"column"} py={6} px={5}>
       <Flex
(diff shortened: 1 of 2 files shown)


packages/extension/src/ui/features/lock/PasswordForm.tsx
```diff
@@ -25,11 +26,9 @@ export const PasswordForm: FC<PasswordFormProps> = ({
   const { errors, isDirty, isSubmitting } = formState
   const { InfoIcon } = icons
 
-  const { error } = useAppState() // FIXME: as a hack we need to use global storage here, as the password form unmounts for the loading screen
```
diegodelrieu, line 28: reading these comments I'm thinking there could be some side effects here, is it all good ?

bluecco: yep, it was an old behaviour

basically it used the appState for loading (so the form was replaced by it)
we changed that in november to use a local loading (now the spinner is in the confirm button) but forgot to remove this part

packages/extension/src/ui/features/lock/LockScreen.tsx
```diff
@@ -63,9 +64,7 @@ export const LockScreen: FC = () => {
                 navigate(target)
                 return true
               } catch {
-                useAppState.setState({
-                  error: "Incorrect password",
-                })
+                setError("Invalid password")
```
simonheys, line 67: Should keep the same wording - 'invalid' implies the password contains unexpected characters
```suggestion
                setError("Incorrect password")
```

packages/extension/src/ui/features/lock/LockScreen.tsx
```diff
@@ -19,6 +18,7 @@ export const LockScreen: FC = () => {
   const navigate = useNavigate()
   const actions = useActions()
   const [isLoading, setIsLoading] = useState<boolean>(false)
+  const [error, setError] = useState<string | undefined>(undefined)
```
simonheys, line 21:
```suggestion
  const [error, setError] = useState<string>()
```

diegodelrieu (approved):

simonheys (approved):

PR #1828, opened 2023-02-23
feat(devnet): allow devs to mint eth

janek26:
Allows for developers to mint ethereum on devnet

https://user-images.githubusercontent.com/6831124/221011444-7a0cefe6-8898-4104-95e1-fc0068fad521.mp4


packages/extension/src/ui/features/funding/FundingScreen.tsx
```diff
@@ -52,6 +55,37 @@ export const FundingScreen: FC = () => {
       <PageWrapper>
         <Title>How would you like to fund your account?</Title>
         <OptionsWrapper>
+          {isDevnet && (
+            <Option
```
diegodelrieu, line 59: why not use swr for this type of query ?

simonheys (approved):

packages/extension/src/ui/features/funding/FundingScreen.tsx
janek26: useMutation is only allowed in swr v2, were still in v1. Closed for now, but lets get a guideline together when to use what

bluecco (approved):

PR #1829, opened 2023-02-24
[BLO-806] Add static analysis with Sonarqube to Argent-X

diegodelrieu:

`Add static code analysis to Argent x, quite a lot to discuss here`

Basically the whole point is to make sure the code quality is checked at every PR and to allow us to enforce best practices whenever new code is shipped. 

The sonarqube dashbaord is available here : https://sonarcloud.io/project/overview?id=argentlabs_argent-x

In sonarqube there's a concept of quality gates which are a set of rules that define what NEW code should go through or not. *(emphasis on new to make sure we can start working on it and having the CI green from the beginning)*. For now they seem a little bit ambitious but we can review them later together and decide what ambitions we have there. Here is the current quality gates (these are company wide expectations - not specific to Argent x)

image

This particular PR fails the quality gates because of PRs that have been merged since then.

### Changes

- Add sonarqube to workflow
- Split CI workflow between unit and e2e tests to allow them to run in parallel
- Extend vite.config.ts to collect test coverage
- Add new script `test:ci` to collect test coverage when running tests in CI . We only collect coverage for UT for now, later we can add coverage for e2e tests but I think it can be a bit misleading.


janek26:
would love @gergold 's feedback on this

gergold (approved):

janek26 (approved):

packages/extension/vite.config.ts
```diff
@@ -15,7 +15,19 @@ export default defineConfig({
       "**/.{idea,git,cache,output,temp}/**",
     ],
     coverage: {
-      exclude: ["**/*.mock.ts", "**/setup.ts"],
+      exclude: [
```
canlopes, line 18: maybe missing _*.spec.*_?

PR #1839, opened 2023-02-27
fix: dont callout voyager on error

janek26:


simonheys (approved):

diegodelrieu (commented):
there's an error that actually says something about voyager somewhere isnt this what this is supposed to tackle ?

dhruvkelawala (approved):

janek26:
> there's an error that actually says something about voyager somewhere isnt this what this is supposed to tackle ?

This error is shown in an ErrorBoundary without any additional checks. So it can be Voyager failing, but can also be different things. 
Voyager reached out to us to change this, as it's not them failing in most cases

PR #1842, opened 2023-02-27
BLO-838 chore: use endpoint to determine token expiry

simonheys:

Can use endpoint to check token expiry (excluding for removal)

### Changes

- change code to use be endpoint to check token expiry

### Checklist

- [x] Code self-reviewed
- [x] Code self-tested

### DIFF
--- a/packages/extension/src/shared/shield/GuardianSignerArgentX.ts
+++ b/packages/extension/src/shared/shield/GuardianSignerArgentX.ts
@@ -1,4 +1,4 @@
-import { GuardianSigner } from "@argent/guardian"
+import { CosignerOffchainMessage, GuardianSigner } from "@argent/guardian"
 import type { CosignerMessage } from "@argent/guardian"
 import type { Signature } from "starknet"
 import { number } from "starknet"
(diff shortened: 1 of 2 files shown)


packages/extension/src/shared/shield/backend/account.ts
```diff
@@ -153,6 +153,16 @@ export const getRegistrationStatus = async (): Promise<RegistrationStatus> => {
   }
 }
 
+export const isTokenExpired = async () => {
+  try {
+    await jwtFetcher(`${ARGENT_API_BASE_URL}/account`)
+    return false
+  } catch (error) {
+    IS_DEV && console.warn(coerceErrorToString(error))
```
diegodelrieu, line 161: I've seen this pattern elsewhere but why collect logs only in dev mode ?

simonheys: The logs can be informative if the service might return unexpected errors as it can here. This has already been useful for debugging with be. Dev only so they don't end up leaking unnecessarily into a prod build.

diegodelrieu (approved):

bluecco (approved):

PR #1844, opened 2023-02-28
[BLO-769] Make wait time dynamic to speed up QA and testing work

diegodelrieu:

` Make wait time dynamic to speed up QA and testing work


packages/extension/.env.example
```diff
@@ -19,3 +19,4 @@ ARGENT_X_STATUS_URL=
 #ARGENT_SHIELD_NETWORK_ID=
 #FEATURE_MULTISIG=
 #FEATURE_VERIFIED_DAPPS=
+#WAIT_TIME=
```
simonheys, line 22: Would prefer a more verbose name - wait time for what?

dhruvkelawala: Agree

dhruvkelawala: It's a common term even used in [ethers.js](https://docs.ethers.org/v5/api/providers/provider/#BaseProvider)

janek26 (changes requested):
does this fix the issue @canlopes is having?
This would require him to always have an "QA" build with an custom env var, not applicable to RCs

I think you should return the inflight transactions from `transactionTracker.update` instead of just a flag which indicates if there are inflight transactions.
Once done, we can calculate that flag in the method plus additionally check if there are any transactions targeting network with id `localhost`. If so, then we can acceleratepolling (aka decrease waiting time between runs)

diegodelrieu:
> does this fix the issue @canlopes is having? This would require him to always have an "QA" build with an custom env var, not applicable to RCs
> 
> I think you should return the inflight transactions from `transactionTracker.update` instead of just a flag which indicates if there are inflight transactions. Once done, we can calculate that flag in the method plus additionally check if there are any transactions targeting network with id `localhost`. If so, then we can acceleratepolling (aka decrease waiting time between runs)

Thanks ! I think this is better now ?

canlopes (approved):

janek26 (approved):

PR #1849, opened 2023-02-28
Develop -> Feature/multisig

dhruvkelawala:


janek26 (approved):

diegodelrieu:
We can just rebase no ?

diegodelrieu (approved):

dhruvkelawala:
We can rebase but I think it's better to use this approach while we move forward, no? Just like a back merge PR, so that we do push directly to master branch.

PR #1852, opened 2023-02-28
feat: transaction tracking for session accounts

janek26:
replaces #1785

### DIFF
--- a/packages/sessions/src/account.ts
+++ b/packages/sessions/src/account.ts
@@ -30,6 +30,8 @@ import {
 
 export class SessionAccount extends Account implements AccountInterface {
   public merkleTree: merkle.MerkleTree
+  public pendingTransactionHashes: string[] = []
+  private pendingExecuteCalls = 0
 
   constructor(
     providerOrOptions: ProviderOptions | ProviderInterface,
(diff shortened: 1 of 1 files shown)


packages/sessions/src/account.ts
```diff
@@ -150,55 +177,72 @@ export class SessionAccount extends Account implements AccountInterface {
...
+      return this.invokeFunction(
         {
+          contractAddress: this.address,
+          calldata,
+          signature,
+        },
+        {
+          maxFee,
+          version,
           nonce,
```
prof197, line 241: I believe this nonce should also be the optimisticNonce, or there will be a signature mismatch.

packages/sessions/src/account.ts
```diff
@@ -150,55 +177,72 @@ export class SessionAccount extends Account implements AccountInterface {
...
-    )
-    const nonce = number.toBN(
-      transactionsDetail.nonce ?? (await this.getNonce()),
-    )
-    let maxFee: number.BigNumberish = "0"
-    if (transactionsDetail.maxFee || transactionsDetail.maxFee === 0) {
-      maxFee = transactionsDetail.maxFee
-    } else {
-      const { suggestedMaxFee } = await this.estimateInvokeFee(
+    this.pendingExecuteCalls += 1
```
prof197, line 180: If this is incremented before the optimisticNonce calculation, it is impossible for the optimisticNonce to match the getNonce() value (which it should when there are no pending executions or open transactions).

PR #1886, opened 2023-03-08
BLO-876: 2fa escape flow

simonheys:

Implement the checks and two-step completion of escape once `triggerEscapeGuardian` period has elapsed.

### Changes

- add checks for new states
- add new screens
- add new transactions
- add storybooks

### Checklist

- [x] Code self-reviewed
- [x] Code self-tested

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -619,7 +619,7 @@
   dependencies:
     "@babel/helper-plugin-utils" "^7.10.4"
 
-"@babel/plugin-syntax-jsx@^7.17.12", "@babel/plugin-syntax-jsx@^7.18.6":
+"@babel/plugin-syntax-jsx@^7.17.12":
   version "7.18.6"
   resolved "https://registry.yarnpkg.com/@babel/plugin-syntax-jsx/-/plugin-syntax-jsx-7.18.6.tgz#a8feef63b010150abd97f1649ec296e849943ca0"
   integrity sha512-6mmljtAedFGTWu2p/8WIORGwy+61PLgOMPOdazc7YoJ9ZCWUyFy3A6CpPkRKLKD1ToAesxX8KGEViAiLo9N+7Q==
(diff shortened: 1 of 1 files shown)


packages/extension/src/shared/shield/GuardianSelfSigner.ts
```diff
@@ -0,0 +1,86 @@
...
+ * such signatures
+ */
+
+export class GuardianSelfSigner extends Signer {
+  public async signMessage(
+    typedData: typedData.TypedData,
+    accountAddress: string,
+  ): Promise<Signature> {
+    const signatures = await super.signMessage(typedData, accountAddress)
+    return [...signatures, ...signatures]
```
diegodelrieu, line 29: does it need to be returned twice ?

diegodelrieu: is it because of the "fake" 2/2 ?

simonheys: Yes it's multisig with same signer twice

packages/extension/src/shared/shield/GuardianSignerArgentX.ts
```diff
@@ -1,15 +1,39 @@
...
+import { isEqualAddress } from "../../ui/services/addresses"
 import { isTokenExpired } from "./backend/account"
 
+/**
+ * 'Escape' entrypoints that can be used when guardian is assigned, but without cosigner
+ */
+export const guardianSignerNotRequired = [
+  "escapeGuardian",
+  "triggerEscapeGuardian",
+]
```
diegodelrieu, line 15: should probably be `as const` since this array is not supposed to be modified

packages/extension/src/ui/features/shield/escape/EscapeGuardianReady.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export const EscapeGuardianReady: FC<EscapeGuardianReadyProps> = ({
+  accountGuardianIsSelf,
+  onRemove,
+}) => {
+  const step = accountGuardianIsSelf ? 2 : 1
+  const title = `Remove Argent Shield now (${step}/2)`
+  const description =
+    step == 1
```
diegodelrieu, line 22: the opposite no ? 1 step = 1 transaction 2 steps = 2 transactions ?

simonheys: Yes it's confusing for sure.
step 1 = tell user they will have to confirm two transactions
step 2 = from user pov one transaction, behind the scenes it's 2 signatures...

diegodelrieu (approved):

packages/extension/src/ui/features/shield/escape/EscapeGuardianReady.tsx
diegodelrieu: ah ok I though this was the number of steps required (1 or 2) got it

bluecco (approved):

PR #1898, opened 2023-03-10
BLO-887 fix: add tab

simonheys:
### Changes

- add tab so it can receive messages

### DIFF
--- a/packages/extension/src/background/activeTabs.ts
+++ b/packages/extension/src/background/activeTabs.ts
@@ -47,7 +47,14 @@ export async function sendMessageToHost(
 ): Promise<void> {
   const tabIds = await getTabIdsOfHost(host)
   const tabs = await activeTabs.get((tab) => tabIds.includes(tab.id))
-  await Promise.allSettled(tabs.map((tab) => tab.port.postMessage(message)))
+
+  for (const tab of tabs) {
+    try {
+      tab.port.postMessage(message)
+    } catch (e) {
+      console.warn(e)
+    }
+  }
 }
 
 export async function sendMessageToActiveTabs(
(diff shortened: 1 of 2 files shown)


packages/extension/src/background/activeTabs.ts
```diff
@@ -47,17 +47,28 @@ export async function sendMessageToHost(
 ): Promise<void> {
   const tabIds = await getTabIdsOfHost(host)
   const tabs = await activeTabs.get((tab) => tabIds.includes(tab.id))
-  await Promise.allSettled(tabs.map((tab) => tab.port.postMessage(message)))
+
+  for (const tab of tabs) {
+    try {
+      tab.port.postMessage(message)
```
janek26, line 53: no await?

packages/extension/src/background/activeTabs.ts
```diff
@@ -47,17 +47,28 @@ export async function sendMessageToHost(
...
   message: MessageType,
 ): Promise<void> {
   const tabs = await activeTabs.get()
 
-  const promises = tabs.map((tab) => tab.port.postMessage(message))
-
-  await Promise.allSettled(promises)
+  for (const tab of tabs) {
+    try {
+      tab.port.postMessage(message)
```
janek26, line 67: same, does it catch errors from within the promise without await?

janek26 (approved):

packages/extension/src/background/activeTabs.ts
simonheys: Surprisingly turns out `port.postMessage` is not a Promise, so it was throwing exception in previous implementation. Actually should handle the error here and disconnect the port - it's usually an 'attempt to use disconnected port object' error

packages/extension/src/background/activeTabs.ts
simonheys: Same

dhruvkelawala (approved):

PR #1909, opened 2023-03-14
[BLO-890] Transaction confirmation screen multisig 

diegodelrieu:

`Implementation of the update threshold and add signers to multisig`

### Changes

- Refacto the multisig screen and form and decouple threshold and signers
- Add messages to the background
- Implement multisig progress bar (split for less than 10 signers, otherwise continuous)


packages/extension/src/background/transactions/transactionMessaging.ts
```diff
@@ -460,6 +462,90 @@ export const handleTransactionMessage: HandleMessage<
     case "TRANSACTION_FAILED": {
       return await actionQueue.remove(msg.data.actionHash)
     }
+
+    case "ADD_MULTISIG_OWNERS": {
+      try {
+        const { address, signersToAdd, newThreshold, currentThreshold } =
+          msg.data
+
+        const thresholdPayload =
```
diegodelrieu, line 471: This part is actually useless and removed in a further PR (no need for threshold payload)

packages/extension/src/ui/features/multisig/MultisigConfirmationsScreen.tsx
```diff
@@ -1,37 +1,142 @@
...
+  const {
+    trigger,
+    formState: { errors },
+    getValues,
+  } = useFormContext<FieldValuesCreateMultisigForm>()
+
   const handleNextClick = () => {
     trigger()
-    // TODO: implement
+    if (!Object.keys(errors).length) {
```
bluecco, line 58: would probably use lodash-es `isEmpty`, feels more clean
```suggestion
    if (isEmpty(errors)) {
```

simonheys (approved):

bluecco (approved):

packages/extension/src/ui/features/multisig/MultisigConfirmationsScreen.tsx
diegodelrieu: thanks !

PR #1936, opened 2023-03-22
BLO-922 fix: 2fa check

simonheys:

2fa should check for both undeployed + upgradable accounts when activating

### Changes

- refactor check
- remove unused hook

### Checklist

- [x] Code self-reviewed
- [x] Code self-tested

### DIFF
--- a/packages/extension/src/ui/features/shield/ShieldAccountStartScreen.tsx
+++ b/packages/extension/src/ui/features/shield/ShieldAccountStartScreen.tsx
@@ -17,10 +17,11 @@ import {
 } from "../../../shared/shield/register"
 import { getVerifiedEmailIsExpiredForRemoval } from "../../../shared/shield/verifiedEmail"
 import { routes } from "../../routes"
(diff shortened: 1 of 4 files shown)


packages/extension/src/ui/features/shield/ShieldAccountStartScreen.tsx
```diff
@@ -80,7 +82,9 @@ export const ShieldAccountStartScreen: FC = () => {
     <NavigationContainer leftButton={<BarBackButton />} title={"Argent Shield"}>
       {isAvailable ? (
         account?.needsDeploy ? (
-          <ShieldAccountNotDeployed />
+          <ShieldAccountNotReady />
+        ) : needsUpgrade ? (
```
diegodelrieu, line 86: no need to do a ternary here I think we can just pass needsUpgrade regardless of it because false or true

simonheys: These are separate flags - would be happy to refactor in a later PR

diegodelrieu (approved):

PR #1944, opened 2023-03-23
Blo 866 update logic of user joining a multisig

dhruvkelawala:

The current issue with joining a multisig account is when we show the public key of the account to be shared for joining the multisig, we create a new account with an address. This newly created account is then showed in the AccountListScreen as multisig with “pending” status.

The problem with this approach is that we cannot later on change the account address with the multisig address. One possible solution that was discussed was to have a separate item in the AccountListScreen which is not of type Account because an Account will always have an address. Instead, we will have a type of item that holds all the BaseMultisigAccount type of data can hold publicKey of that account. We only need PublicKey of the account as everything else will be fetched from Backend. Once we have the data from backend, we can enrich this base account with correct data and convert it into a MultisigAccount type.

### Changes

- Make Account name part of Account
- Remove useAccountMetadata
- Add migration for AccountMetadata to Account for names
- Add usePendingMultisigStore
- Other logical changes I don't remember

### Checklist

- [x] Code self-reviewed
- [x] Code self-tested


dhruvkelawala:
The bug with this PR is that a user with 2 different signers cannot be the part of a same multisig on the same wallet. In other words, if a user creates a "join" multisig and uses the signer from that multisig in "create" multisig, as the address of the multisig is the same, only one will be shown.

packages/extension/src/shared/account/details/updateAccountsWithNames.ts
```diff
@@ -0,0 +1,24 @@
+// Write a function that takes in an array of accounts and returns an array of accounts with the name property added to each account. The name property should be the first and last name of the account holder. The function should be able to handle both individual and business accounts.
```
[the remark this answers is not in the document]
dhruvkelawala: It didn't work. Copilot was too dumb. :)

dhruvkelawala, line 1:
```suggestion
```

packages/extension/src/ui/features/accounts/accountMetadata.state.ts
```diff
@@ -41,41 +47,53 @@ export const useAccountMetadata = create<State>(
...
 export const getAccountName = (
   { address, network }: Account,
-  accountNames: Record<string, Record<string, string>>,
+  accountNames: AccountNames = {},
 ): string => accountNames[network.id]?.[address] || defaultAccountName
 
+/**
+ * @deprecated This is not needed anymore. Name is part of the account object
+ * @param accounts
+ */
```
diegodelrieu, line 63: can we remove all these deprecated things ?

packages/extension/src/ui/features/multisig/MultisigListScreenItem.tsx
```diff
@@ -23,6 +21,17 @@ import { MultisigDeleteModal } from "./MultisigDeleteModal"
 
 const { MoreIcon, ChevronRightIcon } = icons
 
+/**
+ * Same as IAccountListScreenItem, but with account as optional
```
diegodelrieu, line 25: account is not optional here though ?

packages/extension/src/ui/features/multisig/hooks/useMultisigDataForSigner.ts
```diff
@@ -2,57 +2,46 @@ import { useCallback } from "react"
...
-      return
-    }
-
     const data = await fetchMultisigDataForSigner({
-      signer: publicSigner,
-      network: multisigAccount.network,
+      signer: publicKey,
+      network,
     })
+    console.log(
```
diegodelrieu, line 22: can be safely removed i think

packages/extension/src/ui/features/multisig/hooks/usePendingMultsigInfo.ts
diegodelrieu, line 1: typo in the filename

packages/extension/src/ui/services/backgroundMultisigs.ts
```diff
@@ -0,0 +1,44 @@
+import { utils } from "ethers"
+
+import { sendMessage, waitForMessage } from "../../shared/messages"
+import { MultisigData } from "../../shared/wallet.model"
+
+export const createNewMultisigAccount = async (
+  networkId: string,
+  multisigPayload: MultisigData,
+) => {
+  const decodedSigners = multisigPayload.signers.map((signer) =>
```
diegodelrieu, line 10: this should be in the service no ?

packages/extension/src/ui/features/multisig/MultisigListScreenItem.tsx
dhruvkelawala: Yeah I added this comment before actual implementation. Nice catch

packages/extension/src/ui/services/backgroundMultisigs.ts
dhruvkelawala: Wdym?

packages/extension/src/ui/features/multisig/hooks/usePendingMultsigInfo.ts
```diff
@@ -0,0 +1,29 @@
+import { useEffect, useState } from "react"
+
+import { BasePendingMultisig } from "../../../../shared/multisig/types"
+import { pendingMultisigToMultisig } from "../../../../shared/multisig/utils/pendingMultisig"
+import { MultisigWalletAccount } from "../../../../shared/wallet.model"
+import { useMultisigDataForSigner } from "./useMultisigDataForSigner"
+
+export function usePendingMultisigInfo(
```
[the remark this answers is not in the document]
dhruvkelawala: If the Multisig is deployed, that's what it will do. It will check for MultisigData from backend using the signer. If the MultisigData exists, useEffect is triggered which replaces PendingMultisig with Multisig. This hook/file can be removed though because we have it here 

https://github.com/argentlabs/argent-x/blob/35b4583891a227050837aaf8bcccd2bb7bad1a35/packages/extension/src/ui/features/multisig/PendingMultisigListScreenItem.tsx#L29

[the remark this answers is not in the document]
dhruvkelawala: Should we remove it?

packages/extension/src/ui/features/accounts/accountMetadata.state.ts
dhruvkelawala: Let's keep it around for now. Can remove later during the upcoming "refactoring"

packages/extension/src/ui/features/multisig/hooks/usePendingMultsigInfo.ts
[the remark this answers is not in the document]
diegodelrieu: If we use it only here and it's a duplicate then yes

packages/extension/src/ui/services/backgroundMultisigs.ts
diegodelrieu: we dont yet have guidelines on file separation but I guess in the future the handler and the service will be clearly split. no worries for now

PR #1963, opened 2023-03-27
[BLO-905] Remove signer from multisig

diegodelrieu:

`Remove signer from multisig`

### Changes

- Add actions to delete signer (one at a time for now)
- Add delete button in screen


packages/extension/src/background/transactions/transactionMessaging.ts
```diff
@@ -499,6 +499,44 @@ export const handleTransactionMessage: HandleMessage<
...
+
+        const signersPayload = {
+          entrypoint: "removeSigners",
+          calldata: stark.compileCalldata({
+            new_threshold: newThreshold.toString(),
+            signers_to_remove_len: "1",
+            signers_to_remove: utils.hexlify(
+              utils.base58.decode(signerToRemove),
+            ),
+          }),
```
dhruvkelawala, line 515:
```suggestion
          calldata: stark.compileCalldata({
            new_threshold: newThreshold.toString(),
            signers_to_remove: utils.hexlify(
              utils.base58.decode(signerToRemove),
            ),
          }),
```

Don't need to add `*_len`

packages/extension/src/shared/call/changeMultisigSignersCall.ts
```diff
@@ -17,3 +17,21 @@ export const isAddMultisigSignersCall = (
...
+export interface RemoveMultisigSignersCall extends Call {
+  entrypoint: "removeSigners"
+}
+
+export const isRemoveMultisigSignersCall = (
+  call: Call,
+): call is AddMultisigSignersCall => {
+  try {
+    if (call.contractAddress && call.entrypoint === "removeSigners") {
+      validateAndParseAddress(call.contractAddress)
```
dhruvkelawala, line 30:
```suggestion
      validateAndParseAddress(call.contractAddress)
```
This seems better https://github.com/argentlabs/argent-x/blob/bb5d62371cedfa90f6876f207985b74a1d9515c2/packages/extension/src/ui/services/addresses.ts#L76

packages/extension/src/ui/features/multisig/MultisigOwnersScreen.tsx
```diff
@@ -58,15 +68,25 @@ const MultisigOwners = ({ account }: { account: Account }) => {
...
+                  justifyContent="space-between"
+                  alignItems="center"
                 >
                   <H6 color="white">{formatTruncatedSignerKey(signer)}</H6>
-                </Box>
+                  <Button
+                    borderRadius="full"
+                    backgroundColor="neutrals.900"
+                    onClick={() => handleRemoveOwnerClick(signer)}
+                    px="1em"
```
dhruvkelawala, line 85: Why using `em` units? I think it's better to just using `px={1}` or `px={0.5}`

packages/extension/src/shared/wallet.model.ts
```diff
@@ -56,6 +56,12 @@ export type AddOwnerMultisiPayload = {
   currentThreshold?: number
 }
 
+export type RemoveOwnerMultisigPayload = {
+  address: string
+  newThreshold: number
+  signerToRemove: string
+}
+
 export type UpdateMultisigThresholdPayload = {
```
dhruvkelawala, line 65: I don't understand why these types are inside `wallet.model.ts`? They shouldn't be here imo

dhruvkelawala (changes requested):

packages/extension/src/shared/wallet.model.ts
diegodelrieu: true good catch

packages/extension/src/shared/call/changeMultisigSignersCall.ts
diegodelrieu: I saw this in the 2fa calls files, I'm indifferent about either but I think we should only have one way of checking this

simonheys (approved):

dhruvkelawala (approved):

PR #1995, opened 2023-04-04
fix: cleanup deployment logic

janek26:


dhruvkelawala (approved):

gergold (approved):

packages/extension/src/background/keys/keyDerivation.ts
```diff
@@ -27,14 +27,6 @@ export function getPathForIndex(
...
 }
 
-export function getIndexForPath(path: string, baseDerivationPath: string) {
-  if (!path.startsWith(baseDerivationPath)) {
-    throw "path should begin with baseDerivationPath"
-  }
-  const index = path.substring(path.lastIndexOf("/") + 1)
-  return parseInt(index)
-}
-
```
dhruvkelawala, line 37: Let's not remove this method. We are using it internally for multisig

dhruvkelawala (approved):

packages/extension/src/background/keys/keyDerivation.ts
dhruvkelawala, line 37: Let's not remove this method. We are using it internally for multisig

dhruvkelawala (changes requested):

bluecco (approved):

dhruvkelawala (approved):

